PR #2, opened 2019-10-16
Updates the readme to show correct package imports

viljamis:
Noticed that the readme still imports wrong NPM packages in the code examples. This pull request fixes that.


jthoms1:
Thanks for the Update!

PR #7, opened 2019-11-05
fix(): incorrect import paths on windows

schadenn:
Hi everyone,

it's really great for us to have the React output target, saves us a lot of work - so thanks for that. 👌

Currently there is a small bug though. Whenever you're using `path`  and use the return values to generate the import paths in the generated components file, they'll be broken on Windows. `path` will return paths with backslashes on Windows, so import will look like this:
```
import {
  defineCustomElements,
  applyPolyfills
} from "@my\package\dist";
```

I added [slash](https://www.npmjs.com/package/slash) - don't know if you're fine with that or prefer a different solution. Would be great if you'd fix it though.

### DIFF
--- a/packages/angular-output-target/package.json
+++ b/packages/angular-output-target/package.json
@@ -60,7 +60,4 @@
     "testURL": "http://localhost"
   },
   "gitHead": "a3588e905186a0e86e7f88418fd5b2f9531b55e0",
-  "dependencies": {
-    "slash": "^3.0.0"
-  }
 }
--- a/packages/angular-output-target/package.json
+++ b/packages/angular-output-target/package.json
@@ -59,5 +59,5 @@
     ],
     "testURL": "http://localhost"
   },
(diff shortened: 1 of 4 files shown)


jereckley:
You just need to update your stencil projects package.json - 
  "main": "dist/index.js",
  "module": "dist/index.esm.js",
  "types": "dist/types/index.d.ts"

schadenn:
The values are set like that in my package.json.

But I don't think that's got anything to do with the problem/fix I proposed.

jereckley:
Hey! Super sorry for that quick matter of fact comment. Not helpful... I will use more focusing powers next time.

jthoms1:
Slash is actually already included just not as a dependency.
https://github.com/ionic-team/stencil-ds-plugins/blob/master/packages/react-output-target/src/utils.ts#L32

Mind changing the PR to use the internal util function instead?

schadenn:
@jthoms1 

Oh - guess I could've looked there first :)

Did your requested change.

PR #11, opened 2019-11-08
Add null check before registering event handlers

ehaynes99:
Fixes: https://github.com/ionic-team/stencil-ds-plugins/issues/8


jthoms1:
Thank you!

PR #12, opened 2019-11-11
make angular-output-target work with angular production builds

mixer2:
If the angular wrappers are included in an angular application compiled with aot and buildOptimizer, the buildOptimizer will remove the proxyInputs logic since it's not really part of component. It just thinks it's unused code and tries to make the bundle as small as possible.
To prevent the code from getting removed the proxy inputs can be added via ts class annotation. Those are untouched by the build optimizer, as long as they annotate a component class.

### DIFF
--- a/packages/angular-output-target/src/generate-angular-utils.ts
+++ b/packages/angular-output-target/src/generate-angular-utils.ts
@@ -11,14 +11,18 @@ export default async function generateAngularUtils(compilerCtx: CompilerCtx, out
 
 const PROXY_UTILS = `import { fromEvent } from 'rxjs';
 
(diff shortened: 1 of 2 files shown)


misterjojo:
Hi all,

Any idea when this PR will be merged ? I'm facing the same issue on my project, everything works fine until i compile with iot +buildOptimizer.

dopsonbr:
Any update on this?

hypery2k:
any chance to get this merged?

jthoms1:
I believe that this is now obsolute because of PR https://github.com/ionic-team/stencil/pull/2033.

Issue #6, opened 2019-10-24
what is the angularValueAccessorBindings on the new sencil angular plugin?


EstevesDiogo:
`To make use of the AngularOutputPlugin first import it into your stencil.config.ts file. Then add it as an OutputTarget.

import { Config } from '@stencil/core';
import { angularOutputTarget, ValueAccessorConfig } from '@stencil/angular-output-target';

export const config: Config = {
  namespace: 'demo',
  outputTargets: [
    angularOutputTarget({
      componentCorePackage: 'component-library',
      directivesProxyFile: '../component-library-angular/src/directives/proxies.ts',
      valueAccessorConfigs: angularValueAccessorBindings,
    }),
    {
      type: 'dist',
    },
  ],
};`

EstevesDiogo:
Just an update to let any one with the same question to know.
`import { angularOutputTarget, ValueAccessorConfig } from "@stencil/angular-output-target"
​
const EVENTS = {
  "Select": "duetSelect",
  "Change": "duetChange"
};
const ATTRS = {
  "Checked": "checked",
  "Value": "value"
};
const angularValueAccessorBindings: ValueAccessorConfig[] = [
  {
    elementSelectors: [ "duet-checkbox", "duet-toggle" ],
    event: EVENTS.Change,
    targetAttr: ATTRS.Checked,
    type: "boolean"
  },
  {
    elementSelectors: [ "duet-input[type=number]" ],
    event: EVENTS.Change,
    targetAttr: ATTRS.Value,
    type: "number"
  },
  {
    elementSelectors: [ "duet-radio" ],
    event: EVENTS.Select,
    targetAttr: ATTRS.Checked,
    type: "radio"
  },
  {
    elementSelectors: [ "duet-range-slider", "duet-select", "duet-radio-group" ],
    event: EVENTS.Change,
    targetAttr: ATTRS.Value,
    type: "select"
  },
  {
    elementSelectors: ["duet-input:not([type=number])", "duet-textarea"],
    event: EVENTS.Change,
    targetAttr: ATTRS.Value,
    type: "text"
  }
];
​
export const config: Config = {
  outputTargts: [
    angularOutputTarget({
      componentCorePackage: "@duetds/components",
      directivesProxyFile: "../angular/src/directives/proxies.ts",
      valueAccessorConfigs: angularValueAccessorBindings,
    })
    //...
  ]
}`

lidahixson:
[...]

PR #17, opened 2019-11-20
fix(angular): Allow different boolean value accessor attribute

R0muald:
When stared the development of component like radio and checkbox, we used a "selected" attribute to represent the state of the radio/checkbox as boolean.

This PR allows a user to specify on which boolean property they want the nativeElement to be bind to. It is currently forced to the `checked` property.

Part of this PR:
- Added the small documentation related to #6.
- Updated react-output file to make sure both library compiles as expected (based on #15 PR work)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ dist/
 *.swp
 .DS_Store
 .vscode
+yarn.lock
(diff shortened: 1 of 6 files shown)


jthoms1:
Appologizes for not getting to this sooner.  Do you mind updating based on changes that are in master now. I will merge once you do so. Thank you!

jasperbosch:
I ran into the same problem.
Changing `boolean-value-accessor.ts` as suggested by @FrancoisDF fixed just one problem. The same solution should be implemented in `value-accessor.ts`:
```
writeValue(value: any) {
    this.el.nativeElement.<VALUE_ACCESSOR_TARGETATTR> = this.lastValue = value == null ? '' : value;
  }
```
The other adjustments are not necessary in my opinion.

PR #22, opened 2019-11-29
Add 'outputs' to generated proxies.

steindekker:
Add 'outputs' to generated proxies so autocompletion works for events.


jthoms1:
Thanks for the PR, closed in favor of PR #55.

Issue #25, opened 2019-12-17
Missing React component's data attributes 


biirus:
Thank you very much for this incredible work! Stencil is really awesome and allows us to unify our codebase for all projects!

But I noticed that generated React component can't have data attributes. 
I don't know, maybe this is already known issue and there is some plan or roadmap to fix this. 

Could you take a look on my PR, I believe it's able to fix this.
Thank you 🙏

ehaynes99:
This will be very helpful. Data attributes are current swallowed by these wrappers, so components are practically impossible to test with `@testing-library`. JSDom isn't (currently) aware of the text inside of web components, so `getByText` and such don't work. `data-testid` and `aria-label` are the only suitable matchers.

PR #26, opened 2019-12-18
fix(react): pass data- and aria-attributes to React component (#25)

biirus:
Dear Ionic team!
I'm really amazed by you StencilJS and this `react-output-target` helper! But it seems that it misses some important bit. A generated component doesn't pass down `data-` and `aria-` attributes.

It would be great if you'll take a look on this PR.
Thank you 🙏

### DIFF
--- a/packages/react-output-target/react-component-lib/createComponent.tsx
+++ b/packages/react-output-target/react-component-lib/createComponent.tsx
@@ -33,11 +33,17 @@ export const createReactComponent = <PropType, ElementType>(tagName: string) => 
       const { children, forwardedRef, style, className, ref, ...cProps } = this.props;
 
       const propsToPass = Object.keys(cProps).reduce((acc, name) => {
-        if (name.indexOf('on') === 0 && name[2] === name[2].toUpperCase()) {
+        const isEventProp = name.indexOf('on') === 0 && name[2] === name[2].toUpperCase();
+        const isDataProp = name.indexOf('data-') === 0;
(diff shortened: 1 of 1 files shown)


biirus:
Yeah, I think you are totally right. And even more, I found this name `attachEventProps` a bit confusing. It makes more than attaching only events, it attaches all the props, but also syncs events. Maybe it would be better to split this function to two separate.

ehaynes99:
Agreed on the name; it's doing more than that. Essentially those two spots are partitioning the props into those that are passed down as html attributes (during `render`, currently only the event handlers that React magic covers), and those that are assigned to the DOM node directly (during `attachEventProps`). Without the check being mutually exclusive, it will do both. This could potentially fire `@Watch` events twice (though this shouldn't happen in practice, since data attributes should always be strings).

I'm actually a little surprised that it doesn't pass other props that have known attribute names down as attributes as well, but that's beyond the scope of this issue.

petermikitsh:
Hey, I'm interested in getting this PR merged. Let me know if there's anything I can do to help. Thanks!

biirus:
Oh, I'm glad to hear it! 
I'm going to implement Eric's suggestion today or maybe tomorrow.

jthoms1:
Looks good, Thanks!

petermikitsh:
@jthoms1 or @adamdbradley Could you make a new `@stencil/react-output-target` release? Thanks!

PR #41, opened 2020-04-23
feat(angular-output-target): implement setDisabledState in ValueAccessor

WickyNilliams:
The base ValueAccessor for the angular-output-target did not implement [setDisabledState](https://angular.io/api/forms/ControlValueAccessor#setDisabledState). This meant that `disabled` prop of components could not be set/managed through `FormControl`s.

By adding setDisabledState, this will now work:

`app.component.ts`:
```ts
import { Component } from "@angular/core";
import { FormBuilder } from "@angular/forms";

@Component({
  selector: "app-root",
  templateUrl: "./app.component.html",
  styleUrls: ["./app.component.css"],
})
export class AppComponent {
  form = this.formBuilder.group({
	// disabled state will be managed for us by angular
	// and now we can later call `this.form.controls.firstName.enable()`
    firstName: this.formBuilder.control({ value: "", disabled: true }),
    lastName: "",
  });

  constructor(private formBuilder: FormBuilder) {}
}

```

`app.component.html`:
```html
<form [formGroup]="form">
  <my-stencil-input formControlName="firstName" label="First name"> </my-stencil-input>
  <my-stencil-input formControlName="lastName" label="Last name"> </my-stencil-input>

  <button type="submit">Submit</button>
</form>
```

### DIFF
--- a/packages/angular-output-target/resources/control-value-accessors/value-accessor.ts
+++ b/packages/angular-output-target/resources/control-value-accessors/value-accessor.ts
@@ -33,6 +33,6 @@ export class ValueAccessor implements ControlValueAccessor {
   }
 
   setDisabledState(isDisabled: boolean) {
-    this.el.nativeElement.disabled = isDisabled
+    this.el.nativeElement.disabled = isDisabled;
   }
 }
--- a/packages/angular-output-target/resources/control-value-accessors/value-accessor.ts
+++ b/packages/angular-output-target/resources/control-value-accessors/value-accessor.ts
@@ -31,4 +31,8 @@ export class ValueAccessor implements ControlValueAccessor {
   registerOnTouched(fn: () => void) {
     this.onTouched = fn;
   }
+
+  setDisabledState(isDisabled: boolean) {
+    this.el.nativeElement.disabled = isDisabled
+  }
 }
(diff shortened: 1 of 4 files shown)


viljamis:
Yes please, we’d be really happy to see this getting merged in! @adamdbradley Do you see any issues with this change?

viljamis:
@jthoms1 Would you folks need any further help with this and/or more info? This issue is currently forcing our team to use a custom fork of the library.

jthoms1:
Could you add a test for this into the directory 'packages/example-project/component-library-angular/__tests__'.  I will merge once that is in. Thanks!

viljamis:
@jthoms1 We’ll look into this next week.

WickyNilliams:
@jthoms1 I rebased the branch, and added a test.

I ensured the test was testing the correct thing, by first deleting my changes then going through the red/green cycle. There may be a simpler way to test this without using ReactiveForms, but my knowledge of angular is quite limited! So I decided to mirror my example code at the top of the issue in the test.

Let me know if there are any problems!

jthoms1:
Looks good! Thank you!

WickyNilliams:
Thanks!

PR #45, opened 2020-05-12
Fix possible `null` value in attachEventProps

pajter:
`this.ref.current` is possibly `null`. Adding a short-circuit prevents passing `null` to `attachEventProps`.


jthoms1:
With recent changes this code has been rewritten and this issue should no longer occur.  Thanks for the PR!

Issue #54, opened 2020-06-17
Angular event types and docs 


razvangeangu:
The outputs from the angular proxies should inherit the events from the stencil component class.

The current implementation adds the outputs as the properties of the angular class.

```ts
import { Components } from 'component-library'
export declare interface MyButton extends Components.MyButton {}
@ProxyCmp({ inputs: ["buttonType"] })
@Component({
  selector: "my-button",
  changeDetection: ChangeDetectionStrategy.OnPush,
  template: "<ng-content></ng-content>",
  inputs: ["buttonType"],
})
export class MyButton {
  ionFocus!: EventEmitter<CustomEvent>;
  protected el: HTMLElement;
  constructor(c: ChangeDetectorRef, r: ElementRef, protected z: NgZone) {
    c.detach();
    this.el = r.nativeElement;
    proxyOutputs(this, this.el, ["ionFocus"]);
  }
}
```

I believe they should be inherited from the stencil class like this to inherit types and docs in similar fashion to how inputs are added.

```diff
import { Components } from 'component-library'
++ import { Button as IButton } from 'component-library/dist/types/components/my-button/my-button'
++ export declare interface MyButton extends Required<Pick<IButton, 'ionFocus'>> {}
export declare interface MyButton extends Components.MyButton {}
@ProxyCmp({ inputs: ["buttonType"] })
@Component({
  selector: "my-button",
  changeDetection: ChangeDetectionStrategy.OnPush,
  template: "<ng-content></ng-content>",
  inputs: ["buttonType"],
++  outputs: ["ionFocus"]
})
export class MyButton {
--  ionFocus!: EventEmitter<CustomEvent>;
  protected el: HTMLElement;
  constructor(c: ChangeDetectorRef, r: ElementRef, protected z: NgZone) {
    c.detach();
    this.el = r.nativeElement;
    proxyOutputs(this, this.el, ["ionFocus"]);
  }
}
```

jthoms1:
What would be the advantage of making this change. How would it change the experience for consumers of the components?

razvangeangu:
When using the components in Angular it will point to the right interfaces and silent the errors.

<img width="942" alt="Screenshot 2020-06-19 at 23 04 18" src="https://user-images.githubusercontent.com/15063170/85182683-52ecd700-b281-11ea-9290-fdeefd5e574c.png">

jthoms1:
Could you specify the ide you used for the screenshot?

[...]

PR #55, opened 2020-06-17
feat(#54): add outputs with types and docs to angular proxies

razvangeangu:
The outputs from the angular proxies should inherit the events from the stencil component class.

The current implementation adds the outputs as the properties of the angular class.

```ts
import { Components } from 'component-library'
export declare interface MyButton extends Components.MyButton {}
@ProxyCmp({ inputs: ["buttonType"] })
@Component({
  selector: "my-button",
  changeDetection: ChangeDetectionStrategy.OnPush,
  template: "<ng-content></ng-content>",
  inputs: ["buttonType"],
})
export class MyButton {
  ionFocus!: EventEmitter<CustomEvent>;
  protected el: HTMLElement;
  constructor(c: ChangeDetectorRef, r: ElementRef, protected z: NgZone) {
    c.detach();
    this.el = r.nativeElement;
    proxyOutputs(this, this.el, ["ionFocus"]);
  }
}
```

I believe they should be inherited from the stencil class like this to inherit types and docs in similar fashion to how inputs are added.

```diff
import { Components } from 'component-library'
++ import { Button as IButton } from 'component-library/dist/types/components/my-button/my-button'
++ export declare interface MyButton extends Required<Pick<IButton, 'ionFocus'>> {}
export declare interface MyButton extends Components.MyButton {}
@ProxyCmp({ inputs: ["buttonType"] })
@Component({
  selector: "my-button",
  changeDetection: ChangeDetectionStrategy.OnPush,
  template: "<ng-content></ng-content>",
  inputs: ["buttonType"],
++  outputs: ["ionFocus"]
})
export class MyButton {
--  ionFocus!: EventEmitter<CustomEvent>;
  protected el: HTMLElement;
  constructor(c: ChangeDetectorRef, r: ElementRef, protected z: NgZone) {
    c.detach();
    this.el = r.nativeElement;
    proxyOutputs(this, this.el, ["ionFocus"]);
  }
}
```


packages/angular-output-target/src/generate-proxies.ts
```diff
@@ -70,19 +70,23 @@ function getProxy(cmpMeta: ComponentCompilerMeta) {
   if (inputs.length > 0) {
     directiveOpts.push(`inputs: ['${inputs.join(`', '`)}']`);
   }
+  if (outputs.length > 0) {
+    directiveOpts.push(`outputs: ['${outputs.join(`', '`)}']`)
```
razvangeangu, line 74: This is to inform the angular html parser about the available outputs.

packages/angular-output-target/src/generate-proxies.ts
```diff
@@ -70,19 +70,23 @@ function getProxy(cmpMeta: ComponentCompilerMeta) {
...
+  }
 
   const tagNameAsPascal = dashToPascalCase(cmpMeta.tagName);
-  const lines = [`
+
+  const outputsInterface =
+  outputs.length > 0
+    ? `export declare interface ${tagNameAsPascal} extends Required<Pick<JSX.${tagNameAsPascal}, '${outputs.map(o => 'on' + o.replace(/^./, match => match.toUpperCase())).join(`' | '`).replace(/\| $/g, '')}'>> {}`
```
razvangeangu, line 81: Here, the only available import I could find is `JSX.tagNameAsPascal`. Ideally this should come from `dist/types/components/component/component.d.ts`. That way, we can get rid of adding the `on` prefix and correctly inherit the type EventEmitter<WhateverType>.

Pick - will take only the events mentioned in the union
Required - will mark the selected events as required to match the previous implementation.

razvangeangu:
I can see the js and the original ts source absolute paths but I cannot seem to find an easy way to to get the type path.

packages/angular-output-target/src/generate-proxies.ts
```diff
@@ -71,14 +71,15 @@ function getProxy(cmpMeta: ComponentCompilerMeta) {
...
+    directiveOpts.push(`outputs: ['${outputs.map(output => output.name).join(`', '`)}']`)
   }
 
   const tagNameAsPascal = dashToPascalCase(cmpMeta.tagName);
 
+  const typePath = path.relative(componentCorePackage, cmpMeta.sourceFilePath).replace('../src', path.join(componentCorePackage, distTypesDir)).replace('.tsx', '');
   const outputsInterface =
   outputs.length > 0
-    ? `export declare interface ${tagNameAsPascal} extends Required<Pick<JSX.${tagNameAsPascal}, '${outputs.map(o => 'on' + o.replace(/^./, match => match.toUpperCase())).join(`' | '`).replace(/\| $/g, '')}'>> {}`
+    ? `import { ${cmpMeta.componentClassName} as I${cmpMeta.componentClassName} } from '${typePath}';`
```
razvangeangu, line 82: I guess this makes much more sense - however, build is failing due to unresolved types that exist in the global scope of the component-library (i.e. packages/example-project/component-library/src/interfaces.d.ts). Suggestions?

packages/angular-output-target/src/generate-proxies.ts
razvangeangu: I can see how custom names for events would be a problem so I moved from this strategy to the following:

```diff
import { Components } from 'component-library';
import { Button as IButton } from 'component-library/dist/types/components/my-button/my-button'
-- export declare interface MyButton extends Required<Pick<IButton, 'ionFocus'>> {}
export declare interface MyButton extends Components.MyButton {}
@ProxyCmp({ inputs: ["buttonType"] })
@Component({
  selector: "my-button",
  changeDetection: ChangeDetectionStrategy.OnPush,
  template: "<ng-content></ng-content>",
  inputs: ["buttonType"],
  outputs: ["ionFocus"]
})
export class MyButton {
++  /** Emitted after the popover has presented. */
++  ionButtonFocus!: IButton['ionFocus'];
  constructor(c: ChangeDetectorRef, r: ElementRef, protected z: NgZone) {
    c.detach();
    this.el = r.nativeElement;
    proxyOutputs(this, this.el, ["ionFocus"]);
  }
}
```

packages/angular-output-target/src/generate-proxies.ts
```diff
@@ -70,25 +70,36 @@ function getProxy(cmpMeta: ComponentCompilerMeta) {
...
     directiveOpts.push(`inputs: ['${inputs.join(`', '`)}']`);
   }
+  if (outputs.length > 0) {
+    directiveOpts.push(`outputs: ['${outputs.map(output => output.name).join(`', '`)}']`)
+  }
 
   const tagNameAsPascal = dashToPascalCase(cmpMeta.tagName);
-  const lines = [`
+
+  const typePath = path.relative(componentCorePackage, cmpMeta.sourceFilePath).replace('../src', path.join(componentCorePackage, distTypesDir)).replace('.tsx', '');
```
razvangeangu, line 79: replace('../src' and '.tsx') need to go.. Suggestions?

jthoms1:
Thanks for the PR!

jthoms1:
closes #54

Issue #72, opened 2020-06-30
Script compatibility on non *nix systems


MrJean:
I have noticed several things when running the bootstrap command.

* Jest tests fail where the following path is used: Windows doesn't seem to handle the `../../node_modules/.bin/jest` path. This can be fixed by added Jest as a dev dependency to the affected projects and changing `../../node_modules/.bin/jest` to just `jest --passWithNoTests`. Needs to be changed per Lerna package. I did some investigation regarding common dev dependencies with Lerna and found the following (https://github.com/lerna/lerna#common-devdependencies): "Note that devDependencies providing "binary" executables that are used by npm scripts still need to be installed directly in each package where they're used."
* Rimraf is not used in all packages. Since Windows doesn't know `rm -rf` this breaks the build. Needs to be installed per Lerna package. Is the case in component-library-vue and component-library-react.
* The generated component import paths for the Angular output are not right. Output is something like this `component-library\C:\Users\jan.dewilde\projects\stencil-ds-output-targets\packages\example-project\component-library\src\components\my-button\my-button`. I fixed this by changing the way the import path is assembled so this generates `component-library/dist/types/components/my-button/my-button`.

The changed code is from:

```
const typePath = path.parse(path.join(componentCorePackage, cmpMeta.sourceFilePath.replace(path.join(rootDir, 'src'), distTypesDir)));
```

To:

```
  const typePath = path.parse(path.join(componentCorePackage, path.join(cmpMeta.sourceFilePath, '').replace(path.join(rootDir, 'src'), distTypesDir)))
  const importPath = path.join(typePath.dir, typePath.name).replace(/\\/g, "/");
```

I'm willing to open a Pull Request with changes, but it would be great if someone on a *nix system would like to validated the path fix, since I don't have a Mac anymore (and no WSL (yet)).

jthoms1:
Supply a PR and I can test on my Mac. Thanks!

jthoms1:
Actually I think I will include this util from Stencil itself. https://github.com/ionic-team/stencil/blob/6b012a764f5bc8e21372796042970b1f3d77fbe8/src/utils/normalize-path.ts

MrJean:
I'll see if I can implement this. Working on the PR for the other fixes.

MrJean:
[...]

PR #73, opened 2020-06-30
fix(component-library-angular): Component imports in proxies and depe…

MrJean:
…ndency updates for non *nix systems

Reference issue: #72


jthoms1:
Thank you!

MrJean:
Glad I could help :)

PR #75, opened 2020-06-30
allow for component authors to use JSDoc annotation for models

jthoms1:
This PR is to allow component authors to specifiy form based 2 way bindings for Vue and Angular within the JSDoc of their components. This would take the place of 'valueAccessorConfigs' in Angular and 'componentModels' in Vue.

The comment would always go about exposed Events. The following are examples from this repo.

`@bindAttr` (Angular/Vue) requires an attributes that will be bound to. This annotation is for the event that you want to watch for.
`@bindType` (Angular) requires the type that the value should be coerced to and can take an additional value that is a more precise selector. In the case below input has a different value type depending on the input type. (text or number)

```ts
/**
 * Emitted when the value has changed.
 * @bindAttr value
 * @bindType text my-input[type=text]
 * @bindType number my-input[type=number]
 */
@Event() myChange!: EventEmitter<InputChangeEventDetail>
```

```ts
/**
 * Emitted when the checked property has changed.
 * @bindAttr checked
 * @bindType boolean
 */
@Event() myChange!: EventEmitter<CheckboxChangeEventDetail>
```


christian-bromann:
Closing due to inactivity.

PR #78, opened 2020-07-02
docs(general): Added section on how to initiate and use example packages

MrJean:
Based on my work this week with StencilJS and the stencil-ds-ouput-targets repository. I would like to suggest some extra documentation to clarify some small things. Feel free to add remarks/corrections on my writing.

As a next PR I would like to add an example with the my-component component for Angular. If that's OK, please let me know.


jthoms1:
Thank you!

PR #79, opened 2020-07-03
Added example integration of my-component for Angular and documented the use of myCustomEvent in that component

MrJean:


MrJean:
Note: I messed up with amend and reverting ...

jthoms1:
Thanks again!

Issue #81, opened 2020-07-06
bug(angular-output-target): events fire twice


razvangeangu:
Since the events are added to the outputs array - there's no need to create an observable listener anymore to bubble the events. Instead - to keep the outputs array and the types, `fromEvent` can be replaced with an empty `EventEmitter` in the angular proxy utils.

```diff
export const proxyOutputs = (instance: any, el: any, events: string[]) => {
--  events.forEach((eventName) => (instance[eventName] = fromEvent(el, eventName)));
++  events.forEach((eventName) => (instance[eventName] = new EventEmitter()));
};
```

jeserkin:
This needs to be a hotfix thou IMO

dennispidun:
This resolved the issue. Thanks @razvangeangu! 

@jthoms1 please confirm that this pull-request can be merged...I need this to be part of the output-target for angular. Thank you.

khronos90:
This is happening with both Angular 9 and Angular 10, and renders the output completely inusable. Is there any real work around for it? Changing directly the output doesn't make any sense in a CI/CD environment, since it will be regenerated each time the core library is built.

razvangeangu:
The PR should resolve the issue. Alternatively, you may run a custom output target in your own repository or a postbuild script to resolve this issue.

```ts
/**
 * angular-output-target-fix.ts
 *
 * @author @razvangeangu
 */

import { BuildCtx, CompilerCtx, Config, OutputTargetCustom } from '@stencil/core/internal';
import fs from 'fs';
import path from 'path';

const runAngularOutputTargetFix = async (outputTarget: IAngularOutputTarget) => {
  const directivesUtilsPath = path.resolve(__dirname, outputTarget.directivesUtilsFile);

  let directivesUtilsString = await fs.promises.readFile(directivesUtilsPath, { encoding: 'utf-8' });
  directivesUtilsString = directivesUtilsString.replace(
    "import { fromEvent } from 'rxjs';",
    "import { EventEmitter } from '@angular/core';",
  );
  directivesUtilsString = directivesUtilsString.replace(
    'instance[eventName] = fromEvent(el, eventName)',
    'instance[eventName] = new EventEmitter()',
  );

  await fs.promises.writeFile(directivesUtilsPath, directivesUtilsString);
};

interface IAngularOutputTarget {
  directivesUtilsFile: string;
}

export const angularOutputTargetFix = async (
  outputTarget: IAngularOutputTarget,
[...]

PR #82, opened 2020-07-06
fix(#81): add blank event-emitter

razvangeangu:
Since the events are added to the outputs array - there's no need to create an observable listener anymore to bubble the events. Instead - to keep the outputs array and the types, `fromEvent` can be replaced with an empty `EventEmitter` in the angular proxy utils.

```diff
export const proxyOutputs = (instance: any, el: any, events: string[]) => {
--  events.forEach((eventName) => (instance[eventName] = fromEvent(el, eventName)));
++  events.forEach((eventName) => (instance[eventName] = new EventEmitter()));
};
```

### DIFF
--- a/packages/angular-output-target/angular-component-lib/utils.ts
+++ b/packages/angular-output-target/angular-component-lib/utils.ts
@@ -1,4 +1,4 @@
-import { fromEvent } from 'rxjs';
+import { EventEmitter } from '@angular/core';
 
 export const proxyInputs = (Cmp: any, inputs: string[]) => {
   const Prototype = Cmp.prototype;
@@ -25,7 +25,7 @@ export const proxyMethods = (Cmp: any, methods: string[]) => {
 };
 
 export const proxyOutputs = (instance: any, el: any, events: string[]) => {
-  events.forEach((eventName) => (instance[eventName] = fromEvent(el, eventName)));
+  events.forEach((eventName) => (instance[eventName] = new EventEmitter()));
 };
 
 // tslint:disable-next-line: only-arrow-functions


razvangeangu:
Seems that because of some conflicts between stencil's EventEmitter and angular's **subscribe** is not available in angular with the new interfaces and possibly with this blank emitter. Need to investigate this more

TwoToneBytes:
Any update on this? It's a bit annoying have to do `event as CustomEvent<T>` everywhere we use stencil outputs.

razvangeangu:
> Any update on this? It's a bit annoying have to do `event as CustomEvent<T>` everywhere we use stencil outputs.

You would have to add the type anyway to the parameter. No need to use the `as` syntax. You may use it this way:

```ts
handler = (event: CustomEvent<T>) {
    console.log(event.detail); // detail is now `T` and event is `CustomEvent`
}
```

TwoToneBytes:
Unfortunately that doesn't work when you have `strictOutputEventTypes` set to true since the event type that gets exported is coming in as `Event` and you have to explicitly cast it in the event handler.

mrsan22:
Can we get an update on this PR. I was also facing this issue and the solution suggested here seems to solve it.

mrsan22:
@razvangeangu I know you have closed this PR...but I tried your solution with `v0.05` and `@stencil/core v2.0.3` and it seems to fix it.

To your point, on the Angular side, `$event` is also emitting `CustomEvent<T>` as expected from my web component.

Maybe someone from the Stencil team can comment on this. @jthoms1 @adamdbradley any thoughts here.

PR #105, opened 2020-07-31
fix(vue-output-target): only access event.detail.value if it's possible

### LINKED ISSUE #104: [Bug][Vue] TypeError: Cannot read property 'value' of null
The utility should not assume that the event.detail has an object in it
With a custom event where the detail is omitted vue will yield the following error.

`[Vue warn]: Error in v-on handler: "TypeError: Cannot read property 'value' of null"`

Edit:
[CustomEvent](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent) "detail" can be and defaults to null

ghingis:
fixes #104


adamdbradley:
thanks!

PR #110, opened 2020-08-20
Vue: Allow empty string values in common render proxy events

acstll:
Currently, when using Vue's `v-model` on an input, deleting all input's content at once will get you `[object Object]`. This PR should fix that.

image


acstll:
…in the case of a checkbox where the `event.detail.value` can be `false` it also fails, this PR fixes that too.

christian-bromann:
I don't see this code anymore , so I will go ahead and close this.

PR #112, opened 2020-08-21
Added Empty Directive Decorator

himanshugarg574:
Build of component-library-angular is failing with angular 10. This PR fixes it by adding Empty Directive Decorator to value-accessor.ts.


jorgefelico:
Can we get this approved? It's annoying to have to keep going back to every build and manually adding the Directive decorator.

Issue #101, opened 2020-07-23
After upgrade to @stencil/react-output-target 0.0.7 render fails in react


kevinng-staizen:
Hello,

After upgrading @stencil/react-output-target from 0.0.6 to 0.0.7 my stencil components are failing silently and failing to render in React.

Inspecting createComponent.tsx it seems this function has changed quite abit. Please help me by providing some guidance on this issue, thanks much for your time.

Expected: After upgrade from 0.0.6 to 0.0.7 stencil components render in React.

Actual: After upgrade to 0.0.7, stencil components fail to render with no error messages.

Purii:
Fixed my build by setting the introduced options:

```js
reactOutputTarget({
      componentCorePackage: "yourcorepackage",
      proxiesFile: "../proxiesFile.ts",
      includeDefineCustomElements: true,
     // includePolyfills: true // Enable if needed
 }),
```

jdnichollsc:
@Purii what polyfills are included here? I can see an issue with the types for React, pointing to stencil package, is this dependency required now?

Purii:
@jdnichollsc I assume these: https://stenciljs.com/docs/browser-support

jdnichollsc:
Ohh perfect, thanks!

edolix:
It would be awesome to have 0.0.8 out soon 🙏

sentience:
0.0.9 is out, but this is still not fixed.

The `includeDefineCustomElements` and `includePolyfills` options are supposed to default to true, and indeed the code looks intended to do that:

https://github.com/ionic-team/stencil-ds-output-targets/blob/b6c0dab984ec1d2933c52a9a844bc7bd81b7586e/packages/react-output-target/src/plugin.ts#L26-L27

Unfortunately, the `normalizeOutputTarget` function containing that code is only called when validating the plugin's configuration:

https://github.com/ionic-team/stencil-ds-output-targets/blob/b6c0dab984ec1d2933c52a9a844bc7bd81b7586e/packages/react-output-target/src/plugin.ts#L10-L12

…not when actually generating the output:

https://github.com/ionic-team/stencil-ds-output-targets/blob/b6c0dab984ec1d2933c52a9a844bc7bd81b7586e/packages/react-output-target/src/plugin.ts#L13-L16

I think the fix would be:

```js
  async generator(config, compilerCtx, buildCtx) { 
    const timespan = buildCtx.createTimeSpan(`generate react started`, true); 

[...]

PR #118, opened 2020-09-04
fix(react-output): render fails in react

ps011:
fixes https://github.com/ionic-team/stencil-ds-output-targets/issues/101


ps011:
@adamdbradley @manucorporat , Request you to this.

PR #124, opened 2020-11-02
Removed outputs in angular component decorator

khronos90:
Removing the outputs attribute in the component decorator removes the issue with events emitting twice.


denyo:
Related: #81, #126

christian-bromann:
Thanks @khronos90 for raising the PR. It seems like this code has been refactored since you raised it. If the issue still exist, please let me know and let's raise a new issue for it.

PR #126, opened 2020-11-04
fix(#81): change rxjs to event emitter

razvangeangu:
This will make sure the events are fired only once.


jeserkin:
Would @jthoms1 be so kind to review this PR?
Thank you.

nurseiit (approved):

rsurjano:
throwing when compiling with angular 

```javascript
src/directives/angular-component-lib/utils.ts(27,45): error TS6133: 'el' is declared but its value is never read.
```

since `el` parameter is unused in new implementation

christian-bromann:
Fixed in https://github.com/ionic-team/stencil-ds-output-targets/pull/126

Issue #128, opened 2020-11-08
Build is broken since August 2020


hirsch88:
Somehow with the new output for svelte the build is broken.
It requires a `svelte/compile`, but i do not know how we could install that to fix the build.
Or are there any other ideas? happy to help here out.

hirsch88:
It is in the file
https://github.com/ionic-team/stencil-ds-output-targets/blob/master/packages/svelte-output-target/src/output-svelte.ts

```typescript
const svelte = require('svelte/compiler');
```

gnowland:
Duplicate of https://github.com/ionic-team/stencil-ds-output-targets/issues/121

gnowland:
@mihar-22 you worked on this (https://github.com/ionic-team/stencil-ds-output-targets/pull/113) -- would you mind taking a look at this and #121?

To reproduce:
Clone repo, `npm i`, `npm bootstrap`, `npm build` -- fails.

gnowland:
My PR #141 appears to fix the issue for me, can you confirm?

gnowland:
Fixed this by running `npm install svelte --save-dev` in `component-library`... not sure why that'd be necessary though seeing as `svelte` is a dependency of `@stencil/svelte-output-target`...

PR #129, opened 2020-11-08
Fix vue prop syncing and add trailing comma to the accessors of angular

hirsch88:
Hi

First we need to fix this issue #128 .

Vue
With vue we had the issue that the props did update after the first render.

Angular
In the generated accessor files the host list needed a comma as a separator.


OS-jacobbell:
Closing as these bug fixes have been implemented in other PRs.

PR #134, opened 2021-01-15
Use JSX to create component

mykhalov:
This delegates JSX compilation to TypeScript, which can then transform it either to [`React.createElement` or `_jsx` call](https://www.typescriptlang.org/tsconfig#jsx).


christian-bromann:
This is not relevant anymore as we use `@lit/react` for wrapping Stencil components. Thanks anyway for the contribution!

PR #138, opened 2021-02-16
fix(react): Ignore undefined property values

andys8:
## Change

If a property is set to undefined, instead of missing, this can lead to strange behavior in stencil components. For example default values will be overwritten with undefined. This change will filter 'undefined' values from the properties.

## Feedback

Are you open to merging this change?


andys8:
@adamdbradley Merging this, would be awesome :)

christian-bromann:
This is not relevant anymore as we use `@lit/react` now for wrapping the Stencil component. Thanks anyway for the contribution!

PR #148, opened 2021-04-13
fix(react): cleanup events if they are no longer passed as props

WickyNilliams:
Say i have some (react) code like:

```tsx
someCondition 
  ? <MyComponent onSomeEvent={handleSomeEvent} />
  : <MyComponent />
```

Then the event handler doesn’t get removed when `someCondition` is false. 

The current code for syncing props, only considers new props when adding/removing event handlers, and doesn’t remove handlers that existed in previous props but _not_ in new props.

This PR modifies the `attachProps` util to call `removeEventListener` on any listener which is not part of the latest props. 

I had a little trouble writing a good test (read: not testing implementation details), since I think testing-library does not work well with web components. So I just followed the approach in other tests (inspecting `__events`) instead.

I also attempted to fix the build, as the `file://` urls for the svelte output target were causing issues.

### DIFF
--- a/packages/example-project/component-library-react/__tests__/my-component.spec.tsx
+++ b/packages/example-project/component-library-react/__tests__/my-component.spec.tsx
@@ -64,6 +64,25 @@ describe('createComponent - events', () => {
     const attachedEvents = (webcomponent as any).__events;
     expect(Object.keys(attachedEvents)).toContain('myFocus');
   });
+
+  test('should clean up unused events', async () => {
+    const FakeFocus = jest.fn();
+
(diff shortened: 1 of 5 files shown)


viljamis:
Same comment here, would be great to get this merged in.

viljamis:
Any updates to this? We had to create our own version of the React output target to make this work.

christian-bromann:
This patch is not relevant anymore as we have moved to `@lit/react` for wrapping Stencil components within React. Thanks for the contribution.

Issue #123, opened 2020-10-21
[React] Booleans don't get converted correctly 


no-creative-name:
When passing a boolean prop to a wrapped stencil component, I noticed two problems.
`
            <StencilPopup
                isOpen={isPopupOpen}
                stringProp="Yeah"
`
(isPopupOpen is a false boolean)
`
<stencil-popup
    isopen="false"
    stringprop="Yeah"
    string-prop="Yeah"
`

1. The boolean on the resulting stencil component is set to "false" when isPopupOpen is false. In my understanding it should be "true" or "" when true and not set at all when false.
2. String & number props that are set as camelCase in react are added in kebab-case on the resulting stencil component, but boolean props are not. Is this on purpose? Should the props on the react component be passed in camelCase in the first place (which is autosuggested by the TS interface)?

If you lack information, please tell me. I'm happy to provide more. 😊

TRMW:
My understanding based on [this](https://stackoverflow.com/a/31164090) is that React has some logic to conditionally add/remove Boolean attributes on built-in HTML elements such as `input` and `button` (so that if you pass `disabled={false}` the `disabled` attribute is not added to the element at all), but has no such logic for custom element tags, resulting in the behavior described here (`disabled="false"` being added to the element, which is truthy).

It could be helpful for Stencil's React output target to add that translation, since it does know which props are intended to be Booleans.

Thoughts on this? I'd be happy to take a pass at implementing, but could use some logic on the proper place to do so.

TRMW:
Would love to hear any thoughts from those who have contributed most to the React target so far, such as @jthoms1 and @manucorporat. Also thanks for you work on this project!

scabana:
As a workaround, you can alter the output of `attachProps.ts`, replace `if (propType === 'string') {` with 

```
        if (propType === 'boolean') {
          if (newProps[name] === true) {
            node.setAttribute(camelToDashCase(name), camelToDashCase(name));
          } else {
            node.removeAttribute(camelToDashCase(name));
          }
        } else if (propType === 'string') {
```

To see how you can do this: https://github.com/ionic-team/stencil-ds-output-targets/issues/81#issuecomment-721447514

ghaagsma:
Thanks for the workaround @scabana!

[...]

PR #149, opened 2021-04-14
fix(react): improve how props are translated to attributes in html

WickyNilliams:
I noticed that props were not converted to dash-case attributes in the react output target. And also that objects and other non-primitive types would end up in the html like `some-attr="[Object object]"`

This PR fixes the above, and extends the test suite to cover this behaviour.


viljamis:
Would be great to get this fix merged in.

WickyNilliams:
I should add, this seems to improve the load experience when hydrating a pre-rendered page. By passing all the props correctly via HTML, we eliminate a some cumulative layout shift and flickering during hydration. Therefore, I think this is a crucial fix, that should be prioritised!

DarkAtra:
Would be awesome to get this merged soon. I was just trying to get my stencil components to work with react and spent quite a bit of time on the `some-attr="[Object object]"` problem. I was expecting that binding objects and other non-primitive types would [work just fine](https://stenciljs.com/docs/react#bindings).

As a workaround i'm currently using refs but it's kinda ugly:
```ts
const myComponentRef = useRef<HTMLMyComponentElement>(null);
useEffect(() => {
    if (myComponentRef.current) {
        myComponentRef.current.foo = {bar: [1, 2, 3]};
    }
}, [myComponentRef]);

// ...

return <MyComponent ref={myComponentRef}>...</MyComponent>
```

WickyNilliams:
Not sure if it's useful to you, but I actually published a fork with this change and #148 at `@duetds/stencil-react-output-target` until such a time as this gets merged.

DarkAtra:
thanks, i will definitely have a look at it.

packages/example-project/component-library-react/__tests__/my-component.spec.tsx
```diff
@@ -11,25 +11,45 @@ describe('MyComponent', () => {
...
     expect(myComponent.age).toEqual(39);
+    expect(myComponent.getAttribute('age')).toBe('39');
+  });
+
+  it('should get booleans as props and attrs', () => {
+    const { webcomponent: myComponent } = includeWebComponent<HTMLMyInputElement>(
+      renderWithStrictMode(<MyInput clearInput />),
+    );
+    expect(myComponent.clearInput).toEqual(true);
+    expect(myComponent.getAttribute('clear-input')).toBe('true');
```
theo-staizen, line 43: Boolean attributes should have an empty string as a value, not 'true'
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#boolean_attributes

packages/react-output-target/react-component-lib/createComponent.tsx
```diff
@@ -54,14 +55,23 @@ export const createReactComponent = <
...
-            (acc as any)[name] = (cProps as any)[name];
+            acc[name] = value;
           }
         } else {
-          (acc as any)[name] = (cProps as any)[name];
+          // we should only render strings, booleans, and numbers as attrs in html.
+          // objects, functions, arrays etc get synced via properties on mount.
+          const type = typeof value;
+
+          if (type === 'string' || type === 'boolean' || type === 'number') {
```
theo-staizen, line 72: should we be checking if `reflect: true` for each prop?

theo-staizen (commented):
Hi @WickyNilliams, I was looking into this PR and had a couple of questions

packages/react-output-target/react-component-lib/createComponent.tsx
WickyNilliams: I don't think so? Reflect controls whether internal updates to a prop causes a corresponding update in the attr,  right? It doesn't affect how attributes get written from the outside

packages/example-project/component-library-react/__tests__/my-component.spec.tsx
WickyNilliams: I think Stencil might diverge from the spec here. If you set `some-bool="false" ` it will set the underlying value to false IIRC. Perhaps I can be more careful though?

theo-staizen: Hmm, I didn't notice it before, but you are right. That's definitely a bug (or a decision on their part).
If you use the prop `elem.someBool = false` it will correctly remove the attribute, instead of setting its value to `"false"`.

packages/react-output-target/react-component-lib/createComponent.tsx
theo-staizen: i know, but right now if a prop is not reflecting, even if u set it in React, it will update the prop without needing to add an attribute, so why change that?

WickyNilliams: The point is to get as much information into static html as possible, to improve SSR. Any time something is passed as a property you need to wait until after first render, so you get layout shift and flashes of styles as things update

packages/example-project/component-library-react/__tests__/my-component.spec.tsx
WickyNilliams: Perhaps just adding or removing the attr is best. As I said below the aim here is to improve SSR and maximise the amount of data passed via plain ol html

WickyNilliams: One difficulty: booleans props that default to `true` - the only way to set to false is to explicitly pass false

WickyNilliams: The docs show the explicitly passing true/false approach for boolean attributes (I guess they modelled them on aria attributes, which are not like boolean html attributes). So I would say this is intended, not a bug https://stenciljs.com/docs/properties

packages/react-output-target/react-component-lib/createComponent.tsx
theo-staizen: i see, then ignore this one

packages/example-project/component-library-react/__tests__/my-component.spec.tsx
theo-staizen: i think that whoever wrote that part of the docs didn't know or didn't realise he was making a mistake.
they are also showing that u can set a string `false` via JSX, which will result in a type error:
image

WickyNilliams: I think that's because you're setting a property in jsx, whereas here we are dealing with attributes in html (see also: difference between dash case attributes in html and camel case props in jsx). I would lean towards this being intentional, without this there would be no way to set a prop which defaults to true to false in html.

Would be useful for one of the Stencil team to chime in here :)

WickyNilliams: I see what you mean about a mistake in the docs though, as they show the string usage even in jsx! I think the html example is correct, the jsx one is not

TRMW: Just stumbled onto this discussion, and wanted to mention that there's an existing issue (#123) requesting that boolean attributes be omitted from a components' rendered HTML when passed `false`, instead of ending up as the string value `"false"`, as they do now.

Personally, I think it makes sense for Stencil's React wrapper to treat boolean attributes the way React itself treats built-in HTML tags like `input`, where passing `false` removes the attribute.

 Unfortunately we have not heard from anyone on the Stencil team on that thread either.

rwaskiewicz: Hey folks 👋 

First, I apologize for the delay here.  I'm actively looking into #123 and this PR right now.

Second, I've put up a [small PR](https://github.com/ionic-team/stencil-site/pull/826/files) for fixing the docs as pointed out above.  I'm rewriting the `Props` page almost entirely to make it more comprehensive and a little easier to understand. In the mean time, the PR I linked is hopefully a 'quick win' to show folks how you're supposed to use boolean props in a JSX context.

Thirdly, @WickyNilliams can you explain [this comment](https://github.com/ionic-team/stencil-ds-output-targets/pull/149/files#r654983158) to me? I'm not sure I understand the context:

> without this there would be no way to set a prop which defaults to true to false in html.

Are you talking about setting props in a React/JSX context, via HTML? I want to make sure I understand this point so I can take it into consideration.


Thanks!

theo-staizen: @rwaskiewicz if you see my [comment above](https://github.com/ionic-team/stencil-ds-output-targets/pull/149#discussion_r653941740), I explain how the HTML spec defines boolean attributes, which to summarise briefly: `If a boolean attribute is present, its value is true, and if it's absent, its value is false.`

What @WickyNilliams  is pointing out is the case that if a property was set with a default value of `true`, which is something we allow in Stencil, if we follow the spec, we have no way of serialising when the value is set to`false`. The HTML spec probably ignored this and made sure that all HTML elements had props that default to `false`, however this is not something we can guarantee in Stencil.

However, I propose an alternative: we simply assume `default true === true` and `default false === false`, i.e:
- When serializing:
  - We always `include` the attribute when the prop is either `true` or `default true`
  - We always `remove` the attribute when the prop is either `false` or `default false`
- When deserializing:
  - We always treat the existence of the attribute as `true` and set the prop to `true`
  - We always treat the absence of the attribute as `false` and set the prop to `false`

This way we will follow the spec, while allowing for both `default true` and `default false` props to work.

rwaskiewicz (approved):
LGTM! Thanks!

WickyNilliams:
Sorry I missed your earlier messages. Thanks for merging!

rwaskiewicz:
NP! Thanks again!

theo-staizen:
@rwaskiewicz so i take it the whole default boolean attribute debate is benched for now?
I already mentioned in my comment that this PR was causing issues in my code, as it doesn't follow the HTML spec

rwaskiewicz:
@theo-staizen That's correct. Because Stencil doesn't follow the HTML spec, getting it to start to do so would be a pretty large breaking change for folks.  However, I have been looking into multiple issues surrounding prop usage and props don't work the way people expect them to. With that in mind, changing the behavior of props in a future major version of Stencil is something that we're strongly considering.

In the mean time, I've been working on re-vamping the documentation around props to be more explicit about their behavior.  [The PR is open to anyone](https://github.com/ionic-team/stencil-site/pull/832/files) who's interested in taking a look, although it needs a little bit more work before we land it in the docs site.

PR #144, opened 2021-03-19
fix(vue): give name to Vue components

ythomop:
## Pull request checklist

Please check if your PR fulfills the following requirements:
- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [X] Build (`npm run build`) was run locally and any changes were pushed
- [X] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Pull request type

Please check the type of change your PR introduces:
- [X] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?

Issue Number: #143  

## What is the new behavior?
Now, in Vue3-Devtools the components are named correctly.

## Does this introduce a breaking change?

- [ ] Yes
- [X] No

PR #162, opened 2021-06-15
fix(vue): Add name to Vue components

ythomop:
Fixes the generated Vue component to display the actual tag name instead of `AnonymousComponent` in Vue 3 dev tools.

Issue URL: Resolves #144


mwahlhuetter:
This is still an issue with Ionic Framework 6.0.3 and Vue-Devtools 6.0.0 beta 21.
Any update on when this will be merged?

I'm also open to creating a new PR, because there is really just one line that has to change.

sean-perkins:
Thanks for this PR! I am going to close in favor of #257, but will give co-author credit on the commit when it is merged.

PR #170, opened 2021-08-06
fix(vue): only props set by user are passed to web component

liamdebeasi:
See https://github.com/ionic-team/ionic-framework/issues/23539 for more context. In Vue 3.1 there was a breaking change where all properties are now being included in the `props` object even if certain props were not set by the user. These unused props have a value of `undefined`.

With Custom Elements, this causes issues as certain internal values such as event emitters are set on the web component instance resulting in them getting wiped out due to this Vue 3.1 behavior.

This PR fixes the issue by setting the default value of all props to be a `Symbol` so that we can check if the value of a property was not set while respecting properties that can accept `undefined` as a value.

### DIFF
--- a/packages/vue-output-target/vue-component-lib/utils.ts
+++ b/packages/vue-output-target/vue-component-lib/utils.ts
@@ -105,7 +105,7 @@ export const defineContainer = <Props>(
     const navManager: NavManager | undefined = hasRouter ? inject(NAV_MANAGER) : undefined;
     const handleRouterLink = (ev: Event) => {
       const { routerLink } = props;
-      if (!routerLink) return;
+      if (routerLink === EMPTY_PROP) return;
 
       if (navManager !== undefined) {
         let navigationPayload: any = { event: ev };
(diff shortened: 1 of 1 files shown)


packages/vue-output-target/vue-component-lib/utils.ts
```diff
@@ -59,8 +71,8 @@ export const defineContainer = <Props>(
     customElements.define(name, customElement);
   }
 
-  const Container = defineComponent<Props & InputProps>((props, { attrs, slots, emit }) => {
-    let modelPropValue = (props as any)[modelProp];
+  const Container = defineComponent<Props & InputProps>((props: any, { attrs, slots, emit }) => {
```
splitinfinities, line 74: Not blocking, but is there a way to infer/use some typing here, or would the type be way too large - I see it even includes onClick so my assumption is that it could be rather large and hard to add.

splitinfinities (approved):
Nothing blocking, LGTM.

packages/vue-output-target/vue-component-lib/utils.ts
liamdebeasi: Vue uses this `Props` type: https://github.com/vuejs/vue-next/blob/870f2a7ba35245fd8c008d2ff666ea130a7e4704/packages/runtime-core/src/apiDefineComponent.ts#L44

But `PropsOrPropOptions` is just an empty `{}`, though `ExtractPropTypes` looks like this: https://github.com/vuejs/vue-next/blob/870f2a7ba35245fd8c008d2ff666ea130a7e4704/packages/runtime-core/src/componentProps.ts#L106-L109

splitinfinities: Got ya, I checked it out and I see what you mean about `PropsOrPropOptions`. I'm okay with this. Thank you!

Issue #155, opened 2021-05-24
Angular: EventEmitter issue with StrictTemplates set to true


dexster:
Using Angular v12, when setting strictTemplates: true in tsconfig I get this error

`error TS2339: Property 'subscribe' does not exist on type 'EventEmitter<SbgFileUploadChangeEvent>`

The HTML causing the issue looks something like this 
`<my-comp (fileChange)="doSomething()"</my-comp> `

Taking a look at my component during runtime I can see that subscribe is there (see image below)

![Screenshot 2021-05-24 at 08 22 05](https://user-images.githubusercontent.com/6392334/119306112-fbfa3980-bc69-11eb-81ba-3e027dcd470f.png)

Also, If I change the import of EventEmitter in the generated d.ts file for my component from stencil-public-runtime to angular/core this error goes away. Obviously this is not a viable solution though.

The generated proxy is shown below:

```
@ProxyCmp({
    inputs: ['allowedFileTypes', 'disabled', 'fileUploadStatus', 'maxCombinedFileSize', 'maxFileSize', 'multiple', 'required', 'sbgFileUploadStatus'],
    methods: ['reset', 'getSbgFileStatus']
})
@Component({
    selector: 'sbg-file-upload',
    changeDetection: ChangeDetectionStrategy.OnPush,
    template: '<ng-content></ng-content>',
    inputs: ['allowedFileTypes', 'disabled', 'fileUploadStatus', 'maxCombinedFileSize', 'maxFileSize', 'multiple', 'required', 'sbgFileUploadStatus'],
    outputs: ['fileChange']
})
export class SbgFileUploadComponent {
    @Input() disabled: any;

    fileChange!: ISbgFileUpload['fileChange'];
    protected el: HTMLElement;

    constructor(c: ChangeDetectorRef, r: ElementRef, protected z: NgZone) {
        c.detach();
        this.el = r.nativeElement;
        proxyOutputs(this, this.el, ['fileChange']);
    }
}
```

javo2223:
We got the same error when using angular v12?

`Property 'subscribe' does not exist on type 'EventEmitter<any>'.`

Would be very helpful if this bug could be fixed. Our customers can't really use our components in angular v12 applications at the moment.

willmartian:
This issue has been fixed and is available in v0.1.0 of the Angular output target. Thanks for the feedback everyone and thanks to @poulpe38000 for heading the PR contribution.

PR #176, opened 2021-08-24
feat(angular): use Angular's EventEmitter with correct type in generated proxy components

poulpe38000:
* use outputs' complexType definition to detect the output type
* replace className[attribute] value for outputs with Angular's EventEmitter: EvenEmitter<CustomEvent<T>>

The objective is to set the expected output type on each generated proxy component, as the method generateProxyOutputs replaces the original signature by an EventEmitter.

This should fix https://github.com/ionic-team/stencil-ds-output-targets/issues/155

### DIFF
--- a/packages/angular-output-target/src/generate-angular-component.ts
+++ b/packages/angular-output-target/src/generate-angular-component.ts
@@ -40,28 +40,28 @@ export const createComponentDefinition = (
     ),
   );
   const importPath = normalizePath(path.join(typePath.dir, typePath.name));
-  const outputsInterface: string[] = ['']; // Empty first line
+  const outputsInterface: Set<string> = new Set();
 
   const outputReferenceRemap: { [p: string]: string } = {};
   outputs.forEach((output) => {
     Object.entries(output.complexType.references).forEach(([reference, refObject]) => {
       // Add import line for each local/import reference, and add new mapping name
       const remappedReference = `I${cmpMeta.componentClassName}${reference}`;
+      outputReferenceRemap[reference] = remappedReference;
       if (refObject.location === 'local') {
-        outputReferenceRemap[reference] = remappedReference;
-        outputsInterface.push(`import { ${reference} as ${remappedReference} } from '${importPath}';`);
+        outputsInterface.add(`import { ${reference} as ${remappedReference} } from '${importPath}';`);
       } else if (refObject.location === 'import') {
-        outputReferenceRemap[reference] = remappedReference;
         const interfacePath = normalizePath(
           isRelativePath(refObject.path!) ? path.join(typePath.dir, refObject.path!) : refObject.path!,
         );
-        outputsInterface.push(`import { ${reference} as ${remappedReference} } from '${interfacePath}';`);
+        outputsInterface.add(`import { ${reference} as ${remappedReference} } from '${interfacePath}';`);
       }
     });
   });
 
   const lines = [
-    `${outputsInterface.join('\n')}
+    '', // Empty first line
+    `${[...outputsInterface].join('\n')}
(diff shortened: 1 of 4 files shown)


javo2223:
Would this fix the error which is currently there when using angular v12?

`Property 'subscribe' does not exist on type 'EventEmitter<any>'.`

Would be very helpful if this bug could be fixed. Our customers can't really use our components in angular v12 applications at the moment.

poulpe38000:
@javo2223 yes this is the intended fix.

The issue was that the attribute mapping was using Stencil's EventEmitter as a signature. Then, the proxyOutputs method would assign an Angular's EventEmitter.

In strictTemplate mode, the read signature would be Stencil's EventEmitter and not the Angular one, hence the missing properties.

The target here is to keep the event data signature and to assign an Angular EventEmitter instead of the Stencil one.

willmartian:
Hi @poulpe38000, thanks for your PR! I was testing the PR in the Ionic Framework repo and I noticed a couple things:

1. Some imports were repeated when a component would use multiple of the same event detail. I fixed this by storing the imports in a Set rather than an array of strings; committed above. 

2.  For multiple of the component proxies, the input/blur/focus events are referencing undefined types: 
image

packages/angular-output-target/src/generate-angular-component.ts
```diff
@@ -61,7 +75,19 @@ export class ${tagNameAsPascal} {`,
...
+      return type
+        .replace(new RegExp(`^${src}$`, 'g'), `${dst}`)
+        .replace(
+          new RegExp(`([^\\w])${src}([^\\w])`, 'g'),
+          (v, p1, p2) => [p1, dst, p2].join(''),
+        );
+    }, output.complexType.original
+      .replace(/\n/g, ' ')
+      .replace(/\s{2,}/g, ' ')
+      .replace(/,\s*/g, ', '));
```
willmartian, line 89: Could you add a few comments to walk me through your process here?

poulpe38000: Yes, of course: 

The original attribute contains the original type defined by the devs. This could include linebreaks, multiple consecutive spaces, and multiple spaces after commas.

The idea of this is to normalize the original type common issues, by removing linebreaks, replacing consecutive spaces with a single space, and adding a single space after commas.

Do you want me to add more information by adding comments to this part ?

poulpe38000:
> Hi @poulpe38000, thanks for your PR! I was testing the PR in the Ionic Framework [repo](https://github.com/ionic-team/ionic-framework) and I noticed a couple things:
> 
> 1. Some imports were repeated when a component would use multiple of the same event detail. I fixed this by storing the imports in a Set rather than an array of strings; committed above.
> 2. For multiple of the component proxies, the input/blur/focus events are referencing undefined types. Here is an example of the generated proxies in the Ionic Framework repo:
> 
> image

First thing: thanks for the Array to Set, it is indeed cleaner :)

As for the second point, it should not happen, as KeyboardEvent and FocusEvent are supposed to be global imports(set in ```output.complexType.references.location```).

I think the cuplrit here is ```outputReferenceRemap[reference] = remappedReference;```, as it is setting a remap for a global element, and is processed during the remap of the outputs.

packages/angular-output-target/src/generate-angular-component.ts
willmartian: Got it! Added your comment above the regexp. 👍

willmartian:
Ah, it looks like I might have caused that issue above. Your fix solves it. :)

Thanks for this PR, it works great on my end and I think it will be very helpful to other devs--going to merge.

PR #142, opened 2021-03-12
feat(react): compatibility with dist-custom-elements

ianldgs:
Closes #136 

TODO:

- [x] POC
- [ ] Cleanup
- [ ] Unit tests
- [ ] Polyfills?

---

Appreciate any feedback!

Issue #136, opened 2021-02-12
bug: react custom elements does not account for dist-custom-elements-bundle


ianldgs:
With the release of Stencil 2.4.0, we now have a new option for output target, the `dist-custom-elements`.
It will generate a bundle with each component on their own file: 

`class Component extends HTMLElement { ...compiled }`
`customElements.define('component', Component)`-ish

Some Stencil components, like layout components, can cause a lot of cumulative layout shift due to the lazy load.

It would be really nice if `react-output-target` had an option to import from the `dist-custom-elements`, so that bundlers can leverage three-shaking.

I think that it would be as simple as adding an option, and if `true`, add to the react output:

```diff
+import { Component as RawComponent } '${componentCorePackage}/components/component';
export const Component = /*@__PURE__*/createReactComponent<JSX.Component, HTMLComponentElement>('component');
```

The main challenge would be to make the `Component` variable use the `RawComponent` variable, so that static analysis can tell when to include it in the bundle or not.

Any thoughts?

PS: I'd be willing to contribute to a PR, once agreed on how to proceed.

bjankord:
This would be great, for now our team has held off on the react output and I've considered writing my out output target code for React but would love a first class option from the stencil team instead.

krail:
any progess on this?

ianldgs:
@krail I have created #142 but have heard nothing whatsoever from the Stencil team. Tested it with npm workspaces and it works great. The only issue is polyfills. If you need them, you will need to load your entire application's bundle only after they are defined. Can probably be done with `@open-wc/polyfills-loader`.

liamdebeasi:
Thanks for the issue. This has been resolved via https://github.com/ionic-team/stencil-ds-output-targets/pull/182 and will be available in an upcoming release of the React output target.

Arvind6353:
@liamdebeasi Am trying to use the includeImportCustomElements option.

  **reactOutputTarget({
      componentCorePackage: 'fw-design-core',
      proxiesFile: '../react/src/components.ts',
      includeImportCustomElements: true,
      customElementsDir: '????',
    }),**
    
    What exactly should the customElementsDir point to ?
    
    
    My stencil.config.ts config as below :
    

**outputTargets: [
    {
      type: 'dist',
      esmLoaderPath: '../loader',
      copy: [{ src: 'assets' }],
    },
    {
      type: 'dist-custom-elements-bundle',
    },
    {
      type: 'docs-readme',
[...]

PR #182, opened 2021-09-08
feat(react): add support for custom elements build

### LINKED ISSUE #171: dist-custom-elements-bundle support in react-output-target
react-output-target defines all the custom elements by default, when we use "includeDefineCustomElements" as true. Is there a way to not to define everything at once rather per component so that the distributed react components are tree shakable. Is this supported or what's the easiest way to achieve this?

liamdebeasi:
Need to give co-author credit to: https://github.com/ionic-team/stencil-ds-output-targets/pull/142

resolves #171 #136


packages/example-project/component-library-react/src/react-component-lib/createComponent.tsx
```diff
@@ -24,14 +25,16 @@ export const createReactComponent = <
   ExpandedPropsTypes = {}
 >(
   tagName: string,
+  customElement?: any,
```
rwaskiewicz, line 28: Although this is optional, is there any concern that adding it in a place other than the end of the function signature will break existing clients? Or are we planning on making this a breaking change?

packages/react-output-target/src/types.ts
```diff
@@ -5,6 +5,8 @@ export interface OutputTargetReact {
   loaderDir?: string;
   includePolyfills?: boolean;
   includeDefineCustomElements?: boolean;
+  includeImportCustomElements?: boolean;
```
rwaskiewicz, line 8: Is there any documentation we should update to describe these two new fields?

packages/react-output-target/src/output-react.ts
```diff
@@ -68,17 +96,30 @@ import { createReactComponent } from './react-component-lib';\n`;
...
   ];
 
   return final.join('\n') + '\n';
 }
 
-function createComponentDefinition(cmpMeta: ComponentCompilerMeta) {
+/**
+ * Defines the React component that developers will import
+ * to use in their applications.
+ */
```
rwaskiewicz, line 108: Not required - consider adding the `@params` and `@returns` field to this JSDoc

packages/react-output-target/src/output-react.ts
```diff
@@ -68,17 +96,30 @@ import { createReactComponent } from './react-component-lib';\n`;
...
+export function createComponentDefinition(cmpMeta: ComponentCompilerMeta, includeCustomElement: boolean = false) {
   const tagNameAsPascal = dashToPascalCase(cmpMeta.tagName);
+  const importAs = (includeCustomElement) ? tagNameAsPascal + 'Cmp' : undefined;
+
+  let template = `export const ${tagNameAsPascal} = /*@__PURE__*/createReactComponent<${IMPORT_TYPES}.${tagNameAsPascal}, HTML${tagNameAsPascal}Element>('${cmpMeta.tagName}'`;
+
+  if (importAs) {
+    template += `, ${importAs}`;
+  }
```
rwaskiewicz, line 117: Not required, just an alternative
```suggestion
  let template = `export const ${tagNameAsPascal} = /*@__PURE__*/createReactComponent<${IMPORT_TYPES}.${tagNameAsPascal}, HTML${tagNameAsPascal}Element>('${cmpMeta.tagName}'`;

  if (includeCustomElement) {
    template += `, ${tagNameAsPascal}Cmp`;
  }
```

packages/example-project/component-library-react/src/react-component-lib/createComponent.tsx
liamdebeasi: Hmm yes that is true - this would cause a breaking change. The alternative is a bit uglier since you'll need to do `createReactComponent('ion-badge', undefined, undefined, IonBadgeCmp)` but that's probably the correct approach if we are looking to avoid breaking changes.

rwaskiewicz (approved):

PR #200, opened 2021-10-14
feat(angular): add support for custom elements build

willmartian:
The Angular output target can now consume components generated from `dist-custom-elements` output target. To enable this option, set `includeImportCustomElements: true` in the output target config.


packages/angular-output-target/src/generate-angular-component.ts
```diff
@@ -107,14 +119,15 @@ export class ${tagNameAsPascal} {`,
   return lines.join('\n');
 };
 
-function getProxyCmp(inputs: string[], methods: string[]): string {
+//TODO
```
willmartian, line 122:
```suggestion
```
Remove comment

PR #205, opened 2021-11-11
Remove path and os dependencies, lean on Stencil's Systems abstractio…

splitinfinities:
Remove path and os dependencies, lean on Stencil's Systems abstraction to interact with the system.

Depends on https://github.com/ionic-team/stencil/pull/3140

This allows the Angular binding to work with newer versions of Stencil without the need for external dependencies or polyfills.

### DIFF
--- a/packages/angular-output-target/__tests__/output-angular.spec.ts
+++ b/packages/angular-output-target/__tests__/output-angular.spec.ts
@@ -1,6 +1,7 @@
-import { ComponentCompilerMeta } from '@stencil/core/internal';
+import type { ComponentCompilerMeta, Config, CopyTask } from '@stencil/core/internal';
+import type { PackageJSON, OutputTargetAngular } from '../src/types';
 import { generateProxies } from '../src/output-angular';
-import { PackageJSON, OutputTargetAngular } from '../src/types';
+import { createTestingSystem } from './helpers';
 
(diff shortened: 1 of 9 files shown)


packages/angular-output-target/__tests__/output-angular.spec.ts
```diff
@@ -33,14 +42,21 @@ import { Components } from 'component-library';
...
+      {} as any,
+    );
     expect(finalText).toEqual(
       `/* tslint:disable */
 /* auto-generated angular directive proxies */
 import { ChangeDetectionStrategy, ChangeDetectorRef, Component, ElementRef, EventEmitter, NgZone } from '@angular/core';
 import { ProxyCmp, proxyOutputs } from './angular-component-lib/utils';
 
-import { Components } from '../../angular-output-target/dist/types/components';
+import { Components } from '../component-library-angular/dist/types/components';
```
splitinfinities, line 59: This change is worth reviewing. I haven't made a library that doesn't include the component package's name.

splitinfinities: The updated value is actually the correct value - the original test was running from the root of  the angular-output-target path, this one now takes into account the root of the Stencil project (preferred)

Issue #199, opened 2021-10-12
React wrapper treeshaking (dist custom elements) is broken in the Stencil latest release 2.9.0


arvindanta:
Hi Team,

The React wrapper treeshaking ability is broken in the latest release of Stencil 2.9.0

Have added my observation in the below PR.

https://github.com/ionic-team/stencil/pull/3039

https://github.com/ionic-team/stencil/pull/3039#issuecomment-940223140

https://github.com/ionic-team/stencil-ds-output-targets/issues/198#issuecomment-942251290

Kindly help with it. Thanks

liamdebeasi:
Hi there,

We identified an issue with the wrappers when using autoDefineCustomElements that caused treeshaking to not work properly. This was resolved via https://github.com/ionic-team/stencil-ds-output-targets/pull/208.

Issue #198, opened 2021-10-09
Auto import and define dependencies not working with React wrapper - using dist-custom-elements


arvindanta:
Hi Team,

I noticed that the below has been fixed in Stencil .
https://github.com/ionic-team/stencil/pull/3039

I see that auto import and defining dependencies is not working with React wrapper.

If a component uses another component, then importing just the parent component doesn't work. I have to import the dependent components as well to make things work. 

Is there a way to solve this issue?

Thanks

arvindanta:
@splitinfinities 
Please find below the sample stencil component library repo. 
https://github.com/arvindanta/stencil-import-error

https://github.com/arvindanta/stencil-import-error/blob/master/src/components/second-component/second-component.tsx#L15


Here second-component internally uses my-component. Am using react output target.

When I import `import { SecondComponent} from import-error/react`

and use `<SecondComponent />`

my-component contents is not getting loaded . my-component is not getting identified as a web component and only the fallback text is getting displayed.


![Screenshot 2021-10-10 at 10 41 57 PM](https://user-images.githubusercontent.com/91056073/136706212-cdb4fa41-4359-4264-98c7-4ad530a4a8ac.png)

splitinfinities:
I'll circle back soon! This Friday at the latest.

arvindanta:
Thank you. With the dist custom elements (though they are tree shakable) , the dependent components are not getting loaded (auto imported and defined as custom elements) by default. We have to manually load the dependent components and define them. I know there is a latest release 2.9.0 that covers this issue. But this might still not work with React wrapper, as the changes are not made here.

arvindanta:
Hi Team, This is one of the key features I was looking for regarding dist-custom-elements target.

The only issue that this is having is now the React wrapper is not tree shakable.

```
**MyComponent = /*@__PURE__*/ proxyCustomElement(MyComponent, [1, "my-component", {
    "first": [1],
    "middle": [1],
    "last": [1]
  }]);**
```

The above function uses re assign MyComponent variable and this gets exported and is breaking the tree shakable ability (which is very imp with React wrapper(custom elements option)).


Instead of reassigning to MyComponent, If I create a new variable and assign it as below
```
let MyComponent1 = /*@__PURE__*/ proxyCustomElement(MyComponent, [1, "my-component", {
    "first": [1],
    "middle": [1],
    "last": [1]
  }]);**
export { **MyComponent1** as M, defineCustomElement as d };
````
[...]

PR #208, opened 2021-12-01
fix(angular, react, vue): use defineCustomElement import to improve treeshaking

liamdebeasi:
resolves https://github.com/ionic-team/stencil-ds-output-targets/issues/199

When using `autoDefineCustomElements`, Stencil automatically generates a `defineCustomElement()` call in the body of a component file. This is a side effect, which causes bundlers to pull in the entire file when building. When importing components from a single entry point (I.e. `import { IonButton } from '@ionic/vue';`), this causes bundlers to bundle every component with your application, regardless of whether or not the user is actually using the component in their application.

For example, the entry point of @ionic/vue looks something like this:

**index.esm.js**
```js
import { IonAvatar as IonAvatar$1 } from '@ionic/core/component/ion-avatar.js';
import { IonBadge as IonBadge$1 } from '@ionic/core/component/ion-badge.js';

const defineComponent = (tagName, customElement) => {
  // Define Custom Element
  // Define and return the Vue wrapper
}

const IonAvatar = defineComponent('ion-avatar', IonAvatar$1);
const IonBadge = defineComponent('ion-badge', IonBadge$1);

export { IonAvatar, IonBadge }
```

and the file for each component looks something like this (IonAvatar used as example):

```js
import { HTMLElement, proxyCustomElement } from '@stencil/core/internal/client';

let Avatar = class extends HTMLElement {
  ...
};
Avatar = /*@__PURE__*/ proxyCustomElement(Avatar, [33, "ion-avatar"]);
function defineCustomElement$1() {
  if (typeof customElements === "undefined") {
    return;
  }
  const components = ["ion-avatar"];
  components.forEach(tagName => { switch (tagName) {
    case "ion-avatar":
      if (!customElements.get(tagName)) {
        customElements.define(tagName, Avatar);
      }
      break;
  } });
}

const IonAvatar = Avatar;
const defineCustomElement = defineCustomElement$1;

defineCustomElement$1();

export { IonAvatar, defineCustomElement };
```

When bundlers look at `index.esm.js`, they need to evaluate both `ion-avatar.js` and `ion-badge.js`. Calling `defineCustomElement$1();` in each of the component files creates a side effect, resulting in both `ion-avatar` and `ion-badge` being added to an application bundle, even if `ion-avatar` is the only component being used.

------

The wrappers have a secondary defect where if you turn off `autoDefineCustomElements`, child components are never defined because the wrappers only define the parent component.

The wrappers found in this repo do the following to define the components:

```js
if (
    customElement !== undefined &&
    typeof customElements !== 'undefined' &&
    !customElements.get(tagName)
  ) {
    customElements.define(tagName, customElement);
  }
``` 

------

This PR seeks to solve both the wrapper defect and the treeshaking issue by having the wrappers import `defineCustomElement` for each component and call it when the parent component is used.

The one thing to note is that this PR would require the minimum Stencil version to be 2.10.0 as I believe that is the version where the individual `defineCustomElement` functions were generated.

------

Note: This PR does not resolve parts of https://github.com/ionic-team/stencil-ds-output-targets/issues/198 which describes a potential side effect created by `proxyCustomElement`. My initial investigation found that this only impacts Webpack, so Rollup/ESBuild/others should not be impacted by #198.

### DIFF
--- a/packages/angular-output-target/angular-component-lib/utils.ts
+++ b/packages/angular-output-target/angular-component-lib/utils.ts
@@ -45,7 +45,7 @@ export const defineCustomElement = (tagName: string, customElement: any) => {
 // tslint:disable-next-line: only-arrow-functions
 export function ProxyCmp(opts: { defineCustomElementFn?: () => void, inputs?: any; methods?: any }) {
   const decorator = function (cls: any) {
-    const { customElement, inputs, methods } = opts;
+    const { defineCustomElementFn, inputs, methods } = opts;
 
     if (defineCustomElementFn !== undefined) {
       defineCustomElementFn();
--- a/packages/angular-output-target/angular-component-lib/utils.ts
+++ b/packages/angular-output-target/angular-component-lib/utils.ts
@@ -43,11 +43,13 @@ export const defineCustomElement = (tagName: string, customElement: any) => {
 }
 
 // tslint:disable-next-line: only-arrow-functions
-export function ProxyCmp(opts: { tagName: string, customElement?: any, inputs?: any; methods?: any }) {
+export function ProxyCmp(opts: { tagName: string, defineCustomElementFn?: () => void, inputs?: any; methods?: any }) {
   const decorator = function (cls: any) {
     const { tagName, customElement, inputs, methods } = opts;
 
-    defineCustomElement(tagName, customElement);
+    if (defineCustomElementFn !== undefined) {
+      defineCustomElementFn();
+    }
 
     if (inputs) {
       proxyInputs(cls, inputs);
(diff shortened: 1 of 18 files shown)


packages/angular-output-target/angular-component-lib/utils.ts
```diff
@@ -43,11 +43,13 @@ export const defineCustomElement = (tagName: string, customElement: any) => {
 }
 
 // tslint:disable-next-line: only-arrow-functions
-export function ProxyCmp(opts: { tagName: string, customElement?: any, inputs?: any; methods?: any }) {
+export function ProxyCmp(opts: { tagName: string, defineCustomElementFn?: () => void, inputs?: any; methods?: any }) {
   const decorator = function (cls: any) {
     const { tagName, customElement, inputs, methods } = opts;
 
-    defineCustomElement(tagName, customElement);
+    if (defineCustomElementFn !== undefined) {
```
liamdebeasi, line 50: There is another const function called `defineCustomElements` defined above, so that's why I called this `defineCustomElementFn`.

willmartian (approved):
Looks good and makes sense. Thanks for the thorough write-up. 🚀

willmartian (approved):

PR #213, opened 2021-12-16
Add SolidJS Output Target

deleonio:
We love solidjs and stencil - so we working on this plugin.

It works fine - with tiny code.

FYI: @ryansolid


deleonio:
What should I do, too?!

deleonio:
Hi @rwaskiewicz,

what should I do to get the PR through.

My next plan is, to add a Preact plugin.

PR #215, opened 2022-01-03
fix(vue): find correct tag when `ComponentModelConfig.elements` is not an array

yigityuce:
**Output Target**: Vue

**Issue Definition:**
If you have a component with a tag which is included by other component tag name and the "other" component has model config with single string "elements" property, this causes wrongly definition of model config for the first component.

**Example:**
```ts
const componentConfig: ComponentModelConfig[] = [
  {
    elements: 'my-list-item',
    event: 'selectedChange',
    targetAttr: 'selected'
  }
];
```

Lets say I have a model config definition for "my-list-item" component and I have 2 components; "my-list-item" and "my-list".
The current behavior of the autogenerated component definition is:

```ts
export const MyListItem = /*@__PURE__*/ defineContainer<Components.MyListItem>('my-list-item', undefined, [],
'selected', 'selectedChange');

export const MyList = /*@__PURE__*/ defineContainer<Components.MyList>('my-list', undefined, [],
'selected', 'selectedChange');
```

Expected behavior: 

```ts
export const MyListItem = /*@__PURE__*/ defineContainer<Components.MyListItem>('my-list-item', undefined, [],
'selected', 'selectedChange');

export const MyList = /*@__PURE__*/ defineContainer<Components.MyList>('my-list', undefined);
```

**Workaround for current implementation:**
If the user defines the element tag as an array within the model config this fixes the issue by applying a workaround.

```ts
const componentConfig: ComponentModelConfig[] = [
  {
    elements: ['my-list-item'],
    event: 'selectedChange',
    targetAttr: 'selected'
  }
];
```

However the better way is to check the elements property is array and find the model config regarding this.


christian-bromann:
@yigityuce would you be open to revive this PR, would love to get these changes in.

OS-jacobbell (approved):

OS-jacobbell:
This fix makes sense to me. I resolved the merge conflicts and updated it for the current state of the main branch.

PR #216, opened 2022-01-05
fix(react): use createElement named import for better compatibility with Vite

liamdebeasi:
Using `React.createElement` causes Vite to render all components in a Stencil app as `<tagname>` rather than the actual Web Component. See original Ionic issue: https://github.com/ionic-team/ionic-framework/issues/24229

I have identified this to be a Vite bug. See Vite issue: https://github.com/vitejs/vite/issues/6104

According to https://github.com/vitejs/vite/issues/6104#issuecomment-999050698, using the `createElement` named import instead of `React.createElement` fixes the issue. I have verified that this does indeed fix the issue.

I propose we merge this PR in order to get developers unstuck.

### DIFF
--- a/packages/react-output-target/react-component-lib/createComponent.tsx
+++ b/packages/react-output-target/react-component-lib/createComponent.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { createElement } from 'react';
 
 import {
   attachProps,
@@ -80,7 +80,7 @@ export const createReactComponent = <
         style,
       };
 
-      return React.createElement(tagName, newProps, children);
+      return createElement(tagName, newProps, children);
     }
 
     static get displayName() {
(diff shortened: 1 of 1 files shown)


packages/react-output-target/react-component-lib/createComponent.tsx
```diff
@@ -80,7 +80,7 @@ export const createReactComponent = <
         style,
       };
 
-      return React.createElement(tagName, newProps, children);
+      return createElement(tagName, newProps, children);
```
rwaskiewicz, line 83: This feels like it could be easily regressed by someone that doesn't have background knowledge of why we made this change. It may be worthwhile to add a small explainer comment stating we need to use `createElement` in this way + a link to the issue you opened up in the Vite repo

rwaskiewicz (approved):

PR #220, opened 2022-01-15
refactor: updated rimraf dependency

mfranzke:
### Changes

- The breaking change to major version 3 was the following according to the [changelog](https://github.com/isaacs/rimraf/blob/main/CHANGELOG.md#v40):
  > Drop support for Node.js below version 14
  > Remove `glob` dependency entirely. This library now only accepts actual file and folder names to delete.

  (the latter got readded with version 4.2)
- The breaking change to major version 3 was the following according to the [changelog](https://github.com/isaacs/rimraf/blob/main/CHANGELOG.md#v30):
  > Drop support for Node.js below version 6


package.json
```diff
@@ -22,7 +22,7 @@
     "lerna-changelog": "^1.0.1",
     "np": "^6.0.2",
     "prettier": "2.7.1",
-    "rimraf": "^2.6.3",
+    "rimraf": "^4.2.0",
```
rwaskiewicz, line 25: Hey @mfranzke 👋 

Can you do me a favor and run
```
npm install rimraf@4.2.0
```
from the root of this project? CI is currently failing, as this dependency update is not reflected in `./package-lock.json` (causing `npm ci` to fail)

rwaskiewicz (changes requested):

package.json
mfranzke: Hey @rwaskiewicz, yes, sure, I sadly missed that.

rwaskiewicz (approved):
looks good to me - thanks!

Issue #233, opened 2022-03-02
bug: Vue output modelValue is required


markbrockhoff:
Hello,

I noticed that the generated types for Vue always require the property v-model to be defined.
This becomes especially noticeable when using tools like vue-tsc or volar for typechecking a vue project.

![image](https://user-images.githubusercontent.com/95471369/156342151-38dbc8d5-7fcb-4117-86cb-9ddf2da33af0.png)

From my understanding, this comes from the interface `InputProps` defined in the generated file vue-component-lib/utils.ts.
Simply making the property `modelValue` optional seemed to fix the problem in my testing.

```ts
export interface InputProps extends Object {
  modelValue?: string | boolean;
}
```

I already created a small PR to implement the suggested fix: #234

markbrockhoff:
Hi @sean-perkins thanks for the quick response. 
Sure, I created a small demo repo containing the stencil-component-starter, the vue-output-target and a simple vue 3 app showing the problem: https://github.com/markbrockhoff/stencil-component-starter
There's a small readme and a demo bash script that sets everything up for you and reproduces the error.

sean-perkins:
Thanks @markbrockhoff! I'll try to clone and validate this weekend, Monday at the latest.

markbrockhoff:
Is there a date when this fix will be released? It'd be great if we could use it as soon as possible

PR #234, opened 2022-03-02
fix(vue): made modelValue optional

markbrockhoff:
Fixes: #233

### DIFF
--- a/packages/vue-output-target/package-lock.json
+++ b/packages/vue-output-target/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@stencil/vue-output-target",
-  "version": "0.6.0",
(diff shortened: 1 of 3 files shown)


sean-perkins (changes requested):
Confirmed the fix in your reproduction. The optional interface type makes sense.

Can you revert the changes to package.json/package-lock.json? I believe those are auto-versioned through our lerna workspace when releasing. Look at the previous commits to the repo, it looks like maintainers will commit the version change when the package is released.

Thanks!

markbrockhoff:
Thank you! Sure, I just reverted it.

sean-perkins (approved):

PR #237, opened 2022-03-19
chore(): add pull request template

sean-perkins:
Introduces a pull request template for new PRs.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,6 +5,7 @@ Please check if your PR fulfills the following requirements:
 - [ ] Tests for the changes have been added (for bug fixes / features)
 - [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
 - [ ] Build (`npm run build`) was run locally for affected output targets
+- [ ] Tests (`npm test`) were run locally and passed
 
 ## Pull request type
 
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,50 @@
+## Pull request checklist
+
+Please check if your PR fulfills the following requirements:
+
+- [ ] Tests for the changes have been added (for bug fixes / features)
+- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
+- [ ] Build (`npm run build`) was run locally for affected output targets
+
+## Pull request type
+
+<!-- Please do not submit updates to dependencies unless it fixes an issue. -->
+
+<!-- Please try to limit your pull request to one type, submit multiple pull requests if needed. -->
+
+Please check the type of change your PR introduces:
+
+- [ ] Bugfix
+- [ ] Feature
+- [ ] Code style update (formatting, renaming)
+- [ ] Refactoring (no functional changes, no api changes)
+- [ ] Build related changes
+- [ ] Documentation content changes
+- [ ] Other (please describe):
+
+## What is the current behavior?
+
+<!-- Please describe the current behavior that you are modifying. -->
+
+<!-- Issues are required for both bug fixes and features. -->
+
+Issue #:
+
+## What is the new behavior?
+
+<!-- Please describe the behavior or changes that are being added by this PR. -->
+
+-
+-
+-
+
+## Does this introduce a breaking change?
+
+- [ ] Yes
+- [ ] No
+
+<!-- If this introduces a breaking change, please describe the impact and migration path for existing applications below. -->
+
+## Other information
+
+<!-- Any other information that is important to this PR such as screenshots of how the component looks before and after the change. -->
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -29,7 +29,7 @@ Please check the type of change your PR introduces:
 
 <!-- Issues are required for both bug fixes and features. -->
 
-Issue #:
+Issue URL:
 
 ## What is the new behavior?
 


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,50 @@
...
+
+<!-- Issues are required for both bug fixes and features. -->
+
+Issue #:
+
+## What is the new behavior?
+
+<!-- Please describe the behavior or changes that are being added by this PR. -->
+
+-
```
rwaskiewicz, line 37: No action required - FWIW I had been considering changing this to an open text area.

IIRC The original intent was to ensure that folks filled this section out, with the thought
being if we stubbed this section it would encourage people opening PRs to add content
(instead of leaving an empty section).

In practice I've found (personally and in the community) is that folks just delete these and
write a couple sentences.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,50 @@
+## Pull request checklist
+
+Please check if your PR fulfills the following requirements:
+
+- [ ] Tests for the changes have been added (for bug fixes / features)
+- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
+- [ ] Build (`npm run build`) was run locally for affected output targets
+
```
rwaskiewicz, line 8: Should we add an explicit section about running tests? E.g.

```
```

There's some overlap there with 'Tests for the changes have been added (for bug fixes / features)', but I'd also argue there's a case where folks won't add tests, but can still run the existing ones to check for regressions

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,50 @@
...
+- [ ] Refactoring (no functional changes, no api changes)
+- [ ] Build related changes
+- [ ] Documentation content changes
+- [ ] Other (please describe):
+
+## What is the current behavior?
+
+<!-- Please describe the current behavior that you are modifying. -->
+
+<!-- Issues are required for both bug fixes and features. -->
```
rwaskiewicz, line 29: Do we want to make this a requirement? I can see this adding additional noise to the repo where folks add an issue just for the sake of filling out this field

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,50 @@
...
+-
+-
+
+## Does this introduce a breaking change?
+
+- [ ] Yes
+- [ ] No
+
+<!-- If this introduces a breaking change, please describe the impact and migration path for existing applications below. -->
+
```
rwaskiewicz, line 47: Should we ask folks to tell us how they tested the functionality they're proposing? Unit tests can get us there to a certain extent, but I think (for now) a test section would be equally helpful to understand where they're coming from when they submit a PR:

```

## Testing

<!-- Please describe the steps you took to test the changes in this PR. These steps can be programmatic (e.g. unit tests) and/or manual. -->'
```

.github/PULL_REQUEST_TEMPLATE.md
sean-perkins: In Framework, we do this to discourage PRs without a verified/triaged issue. This prevents individuals opening up a PR that does not align with our goals.

We also require a reproduction for issues or a valid use case for a feature, so we have something to validate the PR against. 

It also aids when individuals run into similar problems, they can search the existing issues to find a match vs. PRs do not show up in the issue search.

.github/PULL_REQUEST_TEMPLATE.md
sean-perkins: I think community PRs (the issue attached) should require a reproduction app to validate against. We can then install the dev build for the PR and run the build output to verify anything beyond unit test coverage. This is another thing we do in Framework that works well. 

This reduces the need to test PRs in different ways, based on whatever steps a contributor outlines. 

Thoughts?

.github/PULL_REQUEST_TEMPLATE.md
sean-perkins: If people do that, we will close the PR without review. In Framework we have had the opposite experience (people will fill out the template). I thing a strong hand drives better quality here. If people are unwilling to fill out a 15s template, we are unwilling to review their contribution.

.github/PULL_REQUEST_TEMPLATE.md
rwaskiewicz: I agree with the latter two points here:

> We also require a reproduction for issues or a valid use case for a feature, so we have something to validate the PR against.

⬆️ this one in particular

> It also aids when individuals run into similar problems, they can search the existing issues to find a match vs. PRs do not show up in the issue search.

But I'm not sure I follow the first point

> In Framework, we do this to discourage PRs without a verified/triaged issue. This prevents individuals opening up a PR that does not align with our goals.


Is the discouragement coming from having to open an issue in the first place? IE what's to stop me from putting a PR together, realizing I "need" an issue, add one, then say I met the minimum requirements?  Have y'all found that to be enough?

.github/PULL_REQUEST_TEMPLATE.md
rwaskiewicz: I like it 👍

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,50 @@
...
+- [ ] Documentation content changes
+- [ ] Other (please describe):
+
+## What is the current behavior?
+
+<!-- Please describe the current behavior that you are modifying. -->
+
+<!-- Issues are required for both bug fixes and features. -->
+
+Issue #:
```
amandaesmith3, line 31: Could we do "issue URL" instead like on Framework, to discourage just putting the number instead of a link?

.github/PULL_REQUEST_TEMPLATE.md
amandaesmith3: I think if someone does post an issue just to say they meet the requirements, it makes for a better spot to put discussion on why their issue isn't a good fit, ask for clarification, etc. since other people will be more likely to find it and/or chime in. The author will also be forced to justify their PR more thoroughly by filling out the issue template as well.

sean-perkins: Do we feel aligned on this stance? We can always adapt this template as we receive feedback from the community.

rwaskiewicz: Yeah, I think we're good.  I'm OK with moving forward with this as-is - if we feel we need to, we can adjust the template like @sean-perkins said

rwaskiewicz (approved):

PR #238, opened 2022-03-19
chore(): codeowners format

sean-perkins:
The format for the codeowners file is not registering global owners correctly. The VSCode formatting on markdown contents reformatted `*` to `-` during my previous PR.


rwaskiewicz (approved):
Interesting...I saw this on the first go-round, but GitHub told me "This CODEOWNERS file is valid." 🤷

PR #240, opened 2022-03-19
chore(): add ionic issue bot

sean-perkins:
Adds the Ionitron (ionic issue bot) to the repository to aid with repository management.

### DIFF
--- a/.github/ionic-issue-bot.yml
+++ b/.github/ionic-issue-bot.yml
@@ -48,26 +48,6 @@ lockClosed:
     If this is still an issue with the latest version of the output targets, please create a new issue and ensure the template is fully filled out.
   dryRun: false
 
-stale:
-  days: 365
-  maxIssuesPerRun: 100
-  exemptLabels:
-    - "good first issue"
-    - "triage"
-    - "type: bug"
-    - "type: feature request"
-  exemptAssigned: true
-  exemptProjects: true
-  exemptMilestones: true
-  label: "ionitron: stale issue"
-  message: >
-    Thanks for the issue! This issue is being closed due to inactivity. If this is still
-    an issue with the latest version of the output targets, please create a new issue and ensure the
-    template is fully filled out.
-  close: true
-  lock: true
-  dryRun: false
-
 noReply:
   days: 14
   maxIssuesPerRun: 100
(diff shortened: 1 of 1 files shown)


rwaskiewicz (changes requested):
Two small asks before we land this:

1. We should document (internally) how we expect the Framework + Stencil team should triage issues - what labels to apply/when, etc. I'd be willing to bet there's been some divergence in our respective repository's processes.  That's fine (no complaints there!), but I want to make sure there's a reference point for both teams to understand expectations around handling issues in the repo
2. When we finalize the labels/process described here, we'll need to make sure these labels get added to https://github.com/ionic-team/stencil-ds-output-targets/labels

This looks great though! Thanks sooo much for doing this 🙏

sean-perkins:
Stale section has been removed. I also dropped a notion doc in our internal channel for the initial process.

amandaesmith3 (approved):

.github/ionic-issue-bot.yml
```diff
@@ -0,0 +1,79 @@
...
+        This issue has been labeled as `help wanted`. This label is added to issues
+        that we believe would be good for contributors.
+
+
+        If you'd like to work on this issue, please comment here letting us know that
+        you would like to submit a pull request for it. This helps us to keep track of
+        the pull request and make sure there isn't duplicated effort.
+
+        Thank you!
+    - label: "ionitron: needs reproduction"
```
rwaskiewicz, line 19: Does this label / action conflict with the one below? https://github.com/ionic-team/stencil-ds-output-targets/pull/240/files#diff-5ab3559a040d16f7b7db847060a448dddde18e8320e13b8d5737c31672708375R66

The message associated with this action seems to imply the issue will stay open (so long as they provide a repro), whereas the one below closes the issue. Am I reading this right?

rwaskiewicz:
@sean-perkins I think we're in a good place to add the labels in the PR to https://github.com/ionic-team/stencil-ds-output-targets/labels. Thoughts?

sean-perkins:
Yep, I can add those labels now 👍

.github/ionic-issue-bot.yml
rwaskiewicz: 👍 Makes sense to me - thanks! (serious brain fog day 🙃)

rwaskiewicz (approved):
Approved, Sean's adding the labels now

PR #288, opened 2022-08-22
feat(angular): proxyDeclarationFile to generate proxies file

sean-perkins:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes

N/A - adding the initial API changes for supporting SCAM modules in the Angular output target.

Issue URL: Internal

## What is the new behavior?

- Introduces `proxyDeclarationFile` to the Angular output target config API
- Deprecates `directivesProxyFile` (usage will fallback to this property, if set)

- [x] Yes
- [ ] No

Due to the type change on the `OutputTargetAngular`, `directivesProxyFile` is now optional (up until it is removed) and the new `proxyDeclarationFile` is required. 

Once the SCAM feature is released, developers will need to migrate their stencil config from using `directivesProxyFile` to using the `proxyDeclarationFile` property. Otherwise they will receive a build error when TSC compiles the `stencil.config.ts` file.

## Other information

Note: This PR is merging back into another feature branch.

### DIFF
--- a/packages/angular-output-target/__tests__/plugin.spec.ts
+++ b/packages/angular-output-target/__tests__/plugin.spec.ts
@@ -13,7 +13,7 @@ describe('normalizeOutputTarget', () => {
     }).toThrow(new Error('rootDir is not set and it should be set by stencil itself'));
   });
 
-  it('should throw an error if proxyDeclarationFile is not set', () => {
+  it('should throw an error if proxyDeclarationFile and directivesProxyFile is not set', () => {
     expect(() => {
       normalizeOutputTarget(config, {});
     }).toThrow(new Error('proxyDeclarationFile is required. Please set it in the Stencil config.'));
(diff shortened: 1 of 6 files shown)


rwaskiewicz (approved):

packages/angular-output-target/__tests__/plugin.spec.ts
```diff
@@ -13,19 +13,19 @@ describe('normalizeOutputTarget', () => {
     }).toThrow(new Error('rootDir is not set and it should be set by stencil itself'));
   });
 
-  it('should return fail if proxiesFile is not set', () => {
+  it('should throw an error if proxyDeclarationFile is not set', () => {
```
liamdebeasi, line 16: Since `directivesProxyFile` is deprecated but still supported, should we have separate tests for `directivesProxyFile` and `proxyDeclarationFile` instead?

sean-perkins: I think updating the test description here may make the most sense:
> should throw an error if proxyDeclarationFile and directivesProxyFile is not set

Since the check for throwing an error is dependent on both `directivesProxyFile` and `proxyDeclarationFile` being null/undefined, there is no error behavior to test if only one is defined/undefined.

liamdebeasi (approved):

PR #289, opened 2022-08-22
feat(angular): generate scam modules

sean-perkins:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

N/A

## What is the new behavior?

- Introduces `createSingleComponentAngularModules ` to output target options to generate Angular modules.
- Generates an Angular module for each component with `createSingleComponentAngularModules ` is `true` and `includeImportCustomElements` is `true`.
- Throws an error if trying to generate Angular modules without the custom elements build.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular-output-target/src/generate-angular-modules.ts
+++ b/packages/angular-output-target/src/generate-angular-modules.ts
@@ -1,0 +1,25 @@
+import { dashToPascalCase } from './utils';
+
+const createAngularModuleForComponent = (componentTagName: string) => {
+  const tagNameAsPascal = dashToPascalCase(componentTagName);
+
+  const moduleDefinition = `
+@NgModule({
+  declarations: [${tagNameAsPascal}],
+  exports: [${tagNameAsPascal}]
+})
+export class ${tagNameAsPascal}Module {}
+`;
+
+  return moduleDefinition;
+};
+
+/**
+ * Creates an Angular module for each component.
+ * @param componentTagNames The tag names of the components to create modules for.
+ * @returns An array of Angular modules.
+ */
+export const generateAngularModules = (componentTagNames: string[]) => {
+  const modules = componentTagNames.map(createAngularModuleForComponent);
+  return modules;
+};
--- a/packages/angular-output-target/src/generate-angular-modules.ts
+++ b/packages/angular-output-target/src/generate-angular-modules.ts
@@ -1,5 +1,10 @@
 import { dashToPascalCase } from './utils';
 
+/**
+ * Creates an Angular module declaration for a component wrapper.
+ * @param componentTagName The tag name of the Stencil component.
+ * @returns The Angular module declaration as a string.
+ */
 const createAngularModuleForComponent = (componentTagName: string) => {
   const tagNameAsPascal = dashToPascalCase(componentTagName);
 
@@ -20,6 +25,5 @@ export class ${tagNameAsPascal}Module {}
  * @returns An array of Angular modules.
  */
 export const generateAngularModules = (componentTagNames: string[]) => {
-  const modules = componentTagNames.map(createAngularModuleForComponent);
-  return modules;
+  return componentTagNames.map(createAngularModuleForComponent);
 };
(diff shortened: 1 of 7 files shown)


packages/angular-output-target/src/types.ts
```diff
@@ -18,6 +18,10 @@ export interface OutputTargetAngular {
   excludeComponents?: string[];
   includeImportCustomElements?: boolean;
   customElementsDir?: string;
+  /**
+   * `true` to generate a single component Angular module for each component.
+   */
+  createAngularModules?: boolean;
```
rwaskiewicz, line 24: I wonder if it makes sense to rename this to be a little more specific to single-component modules. Thoughts on renaming this to 
```suggestion
  createSingleComponentAngularModules?: boolean;
```

packages/angular-output-target/src/utils.ts
```diff
@@ -85,6 +85,15 @@ export async function readPackageJson(config: Config, rootDir: string) {
   return pkgData;
 }
 
+/**
+ * Creates an import statement for a list of named imports from @angular/core.
+ * @param imports The list of named imports.
+ * @returns The import statement as a string.
+ */
+export const createAngularCoreImportStatement = (imports: string[]) => {
+  return `import { ${imports.join(', ')} } from '@angular/core';`;
```
rwaskiewicz, line 94: Can we add some protection in the event that `imports` has a length of zero (and a test for it)? Otherwise, I think ATM this generates invalid TS:
```
`import {  } from '@angular/core';`
```

packages/angular-output-target/src/generate-angular-modules.ts
```diff
@@ -0,0 +1,25 @@
...
+};
+
+/**
+ * Creates an Angular module for each component.
+ * @param componentTagNames The tag names of the components to create modules for.
+ * @returns An array of Angular modules.
+ */
+export const generateAngularModules = (componentTagNames: string[]) => {
+  const modules = componentTagNames.map(createAngularModuleForComponent);
+  return modules;
```
rwaskiewicz, line 24: I don't know if there's follow up work to be done that will make this comment moot, but consider inlining the return here:
```suggestion
  return componentTagNames.map(createAngularModuleForComponent);
```

packages/angular-output-target/src/generate-angular-modules.ts
```diff
@@ -0,0 +1,25 @@
+import { dashToPascalCase } from './utils';
+
+const createAngularModuleForComponent = (componentTagName: string) => {
```
rwaskiewicz, line 3: Minor, can we add a JSDoc for this fn?
```suggestion
/**
 * Generate the code for a single Angular module
 * @param componentTagName the tag name of the Stencil component to generate the module for
 * @returns the generated module definition
 */
const createAngularModuleForComponent = (componentTagName: string) => {
```

packages/angular-output-target/src/output-angular.ts
```diff
@@ -108,6 +130,18 @@ import { ProxyCmp, proxyOutputs } from './angular-component-lib/utils';\n`;
     sourceImports = cmpImports.join('\n');
   }
 
+  const proxyModules: string[] = [];
+
+  if (createAngularModules) {
+    // Generating Angular modules is only supported in the dist-custom-elements build
+    if (outputTarget.includeImportCustomElements === false) {
```
rwaskiewicz, line 137: What if `includeImportCustomElements` is elided from the config? Should we fail there as well?

rwaskiewicz (changes requested):

rwaskiewicz (approved):

liamdebeasi (approved):
The code looks good. Were we able to verify whether or not having the module + component in the same file works well with webpack's treeshaking?

sean-perkins:
@liamdebeasi not yet, that is my goal to get to by EOD tomorrow. My plan is if webpack isn't able to perform treeshaking on a single entry point, I'll still use the boilerplate here for module creation, but generate a new file for each component and module instead.

My suspicion is that we will have to end up generating a directory structure like:
```
component-lib/
  src/
      button/
           button.component.ts
           button.module.ts
           index.ts
      card/
           card.component.ts
           card.module.ts
           index.ts
```

and that developers will need to import from:
```ts
import { ButtonModule } from 'component-lib/button';
import { CardModule } from 'component-lib/card';
```

Similar to how Material Angular works: https://material.angular.io/components/button/api.

Otherwise they would have had a single entry point for DX.

PR #298, opened 2022-09-02
feat(angular-output-target): generate single component angular modules

sean-perkins:
## Pull request checklist

Please check if your PR fulfills the following requirements:

- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally for affected output targets
- [x] Tests (`npm test`) were run locally and passed
- [x] Prettier (`npm run prettier`) was run locally and passed

## Pull request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

Issue URL: #207

## What is the new behavior?

- Developers can generate single component Angular modules with the `createSingleComponentAngularModules` option in the output target options (developer preview feature - experimental)

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

Dev-build: `0.6.1-dev.11676048291.1d2fefc9`

PR #302, opened 2022-09-30
refactor(angular-output-target): split up component proxy code generation

sean-perkins:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes

N/A

## What is the new behavior?

- Adds test coverage to code generation
- Splits up proxy generation into separate functions

- [ ] Yes
- [x] No

## Other information

Split from: https://github.com/ionic-team/stencil-ds-output-targets/pull/298


packages/angular-output-target/src/generate-angular-component.ts
```diff
@@ -1,132 +1,165 @@
...
+   * replacing consecutive spaces with a single space, and adding a single space after commas.
+   */
+  return Object.entries(event.complexType.references)
+    .filter(([_, refObject]) => refObject.location === 'local' || refObject.location === 'import')
+    .reduce(
+      (type, [src, dst]) => {
+        const renamedType = `I${componentClassName}${type}`;
+        return renamedType
+          .replace(new RegExp(`^${src}$`, 'g'), `${dst}`)
+          .replace(new RegExp(`([^\\w])${src}([^\\w])`, 'g'), (v, p1, p2) => [p1, dst, p2].join(''));
```
rwaskiewicz, line 101: FWIW, I think this can be simplified to:
```suggestion
          .replace(new RegExp(`(\W)${src}(\W)`, 'g'), (v, p1, p2) => [p1, dst, p2].join(''));
```

rwaskiewicz, line 101: We might want to consider adding a one-liner comment here to explain what this does (or at least its intent):
```suggestion
          // capture all instances of the `src` field surrounded by non-word characters on each side, join them
          .replace(new RegExp(`([^\\w])${src}([^\\w])`, 'g'), (v, p1, p2) => [p1, dst, p2].join(''));
```

rwaskiewicz (approved):
LGTM all in all. This is all code I previously approved, and I don't love the idea of double-jeopardy when it comes to code reviews - I left a few comments, but consider them non-blocking/as suggestions

liamdebeasi (approved):

packages/angular-output-target/src/generate-angular-component.ts
sean-perkins: I'm going to opt against taking this commit for this PR. If you want to open up another PR to make that isolated change, I can review it 👍 that way in the event of a bug around this code in the future, it will be easier to reason about the timeline of changes.

Issue #212, opened 2021-12-08
Output events not rendered correctly


adoumas:
Hello,

I try to use the AngularOutputTargets and I spot an issue when you use the custom eventName prop on components.

`@Event({eventName: 'mdc-input'}) mdcInput: EventEmitter<KeyboardEvent>;`

The output target creates the wrong output definition because when you use 'kebab' case event names you should wrap them in quotes so can pass the compiler validation.

`
export declare interface MdcTextField extends Components.MdcTextField {
  /**
   * Emitted when a keyboard input occurred. 
   */
  mdc-input: EventEmitter<CustomEvent<any>>;
}
`

And I spot the issue in this file '/src/generate-angular-component.ts' line 82. My suggestion is to replace the output name with quotes.

from
`componentEvents.push(`  ${output.name}: EventEmitter<CustomEvent<${outputTypeRemapped.trim()}>>;`);`
to
`componentEvents.push(`  '${output.name}': EventEmitter<CustomEvent<${outputTypeRemapped.trim()}>>;`);`

Thank you guys

adoumas:
@sean-perkins thank you m8. waiting to be merged. Cheers!

PR #307, opened 2022-10-24
fix(angular-output-target): escape kebab-case event names

sean-perkins:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Kebab-case event names do not generate proxies correctly. 

Issue URL: #212

## What is the new behavior?

- Kebab-case event names are escaped, allowing the angular output target to correctly generate the component proxies.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular-output-target/src/generate-angular-component.ts
+++ b/packages/angular-output-target/src/generate-angular-component.ts
@@ -150,7 +150,7 @@ export const createComponentTypeDefinition = (
   const eventTypes = publicEvents.map((event) => {
     const comment = createDocComment(event.docs);
     let eventName = event.name;
-    if (event.name.indexOf('-') !== -1) {
+    if (event.name.includes('-')) {
       // If an event name includes a dash, we need to wrap it in quotes.
       // https://github.com/ionic-team/stencil-ds-output-targets/issues/212
       eventName = `'${event.name}'`;
--- a/packages/angular-output-target/src/generate-angular-component.ts
+++ b/packages/angular-output-target/src/generate-angular-component.ts
@@ -149,8 +149,14 @@ export const createComponentTypeDefinition = (
   });
   const eventTypes = publicEvents.map((event) => {
     const comment = createDocComment(event.docs);
+    let eventName = event.name;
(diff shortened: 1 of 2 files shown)


tanner-reits (approved):
This looks fine, but is there any harm in just wrapping these in strings by default?

sean-perkins:
@tanner-reits there would be no harm in always escaping the type name in the interface declaration. I think a reason we wouldn't always escape is to align closer to how a developer would write an interface manually. Formatters such as prettier remove quotes around type declarations that do not require escaping.

If we landed on wanting to always escape, we would likely want to get larger team exposure, as it would change the generated output of existing projects using the Angular output target.

packages/angular-output-target/src/generate-angular-component.ts
```diff
@@ -149,8 +149,14 @@ export const createComponentTypeDefinition = (
   });
   const eventTypes = publicEvents.map((event) => {
     const comment = createDocComment(event.docs);
+    let eventName = event.name;
+    if (event.name.indexOf('-') !== -1) {
```
rwaskiewicz, line 153: No action required, this could be shortened to
```suggestion
    if (event.name.includes('-')) {
```

rwaskiewicz (approved):

Trendy:
> @tanner-reits there would be no harm in always escaping the type name in the interface declaration. I think a reason we wouldn't always escape is to align closer to how a developer would write an interface manually. Formatters such as prettier remove quotes around type declarations that do not require escaping.
> 
> If we landed on wanting to always escape, we would likely want to get larger team exposure, as it would change the generated output of existing projects using the Angular output target.

Would it make more sense to merge this to fix the existing issue, and consider escaping all the definitions as a potential enhancement?

iron2414mw:
Any updates on this? If not, could you guys merge this please?

Trendy:
> Any updates on this? If not, could you guys merge this please?

I'll second this. Currently, we're manually applying this patch to our repositories to get around the issue with kebab case event names.

Issue #303, opened 2022-10-20
bug: Event sncStateChange is not emitted by any applicable directives nor by snc-navigation-sidebar element


ofirrifo:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil-ds-output-targets/issues) that already report this problem, without success.

### Stencil Version

"@stencil/core": "2.17.1",

### Stencil Framework Output Target

Angular

### Stencil Framework Output Target Version

"@stencil/angular-output-target": "^0.4.0",

### Current Behavior

First I want to mention I read [this](https://github.com/ionic-team/stencil-ds-output-targets/issues/54) 
Which is the same issue as I have but despite I'm using the latest version I still have the issue.

When I'm using the components in Angular it show me this error.
`Event sncStateChange is not emitted by any applicable directives nor by snc-navigation-sidebar element`

For this code
```
  <snc-navigation-sidebar
    (sncStateChange)="doSomthing($event)">
  </snc-navigation-sidebar>
```

And I found that the outputs from the angular proxies is not exist as part of the @Component decorator.

```ts
@ProxyCmp({
  defineCustomElementFn: undefined,
  inputs: ['customContent', 'isOpen', 's1PagesData', 'state']
})
@Component({
  selector: 'snc-navigation-sidebar',
  changeDetection: ChangeDetectionStrategy.OnPush,
  template: '<ng-content></ng-content>',
  inputs: ['customContent', 'isOpen', 's1PagesData', 'state'],
})
export class SncNavigationSidebar {
  protected el: HTMLElement;
  constructor(c: ChangeDetectorRef, r: ElementRef, protected z: NgZone) {
    c.detach();
    this.el = r.nativeElement;
    proxyOutputs(this, this.el, ['sncStateChange']);
  }
}
```

### Expected Behavior

To have the `outputs: ["anyOutpout"]` as part of the @Component 

### Steps to Reproduce

- create an stenciljs component with event
- create angular component from stenciljs project
- create a new angular project and try to use the angular component output

### Code Reproduction URL

[...]

Issue #219, opened 2022-01-11
Angular binding error : Argument of type 'Event' is not assignable to parameter of type CustomEvent


viico:
Hi,

We use Stencil to build a web component library. We follow the stencil's guide Angular Integration (https://stenciljs.com/docs/angular) in order to create an angular library to use our web components in an angular project.

In the target project, with `strictTemplates` key at true, we can't use the generic type `CustomEvent` for web component events in our application.

The error is :
```
Error: src/app/app.component.html:1:74 - error TS2345: Argument of type 'Event' is not assignable to parameter of type 'CustomEvent<NameClickedEvent>'.
  Type 'Event' is missing the following properties from type 'CustomEvent<NameClickedEvent>': detail, initCustomEvent

1 <my-component first="Stencil" last="Compiler" (nameClicked)="nameClicked($event)"></my-component>
```

We pushed a minimal repository to reproduce the error :
1. clone the repo https://github.com/viico/angular-output-target-custom-event
2. `npm install` and `npm run build` in stencil-library folder
3. `npm install` and `npm run build` in angular-workspace folder
4. `npm install` and `ng serve` in my-app folder => errors occurs during ng serve

It works if we set `strictTemplates` at false or if we use `any` or `Event` type but these solutions are not feasible.

```typescript
//nameClicked($event: any) { // works
nameClicked($event: CustomEvent<NameClickedEvent>) { // fails
  console.log($event);
}
```

After a 2 day search we don't see what we did wrong, could you help us to find the problem ?

Thanks

viico:
I forgot an information.

When I manually update the auto-generated angular directive proxies file to add the `outputs` key in the `Component` decorator, it fixes the error, but the event is trigger twice (relative to this PR https://github.com/ionic-team/stencil-ds-output-targets/pull/202).

```typescript
@Component({
  selector: 'my-component',
  changeDetection: ChangeDetectionStrategy.OnPush,
  template: '<ng-content></ng-content>',
  inputs: ['first', 'last', 'middle'],
  outputs: ['nameClicked'] // fix the error, but event is trigger twice
})
```

mmacret:
Any updates on this? it is really making it hard to use the Angular wrappers...

tfrijsewijk:
[...]

PR #308, opened 2022-11-03
fix(angular-output-target): type resolution with strict templates

sean-perkins:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Angular generated output bindings do not correctly detect the custom event types.

Issue URL: #303, #219

## What is the new behavior?

- Clears the Angular metadata for each generated `@Output()`
  - This prevents Angular from creating its own event listener/subject (prevents duplicate event emissions)
  - This adds support for custom events to be registered with Angular Language Service
- Reworked how the text output is created. Maintaining spacing and line breaks became too complex with using string interpolation

- [ ] Yes
- [x] No

## Other information

This PR supersedes https://github.com/ionic-team/stencil-ds-output-targets/pull/266. The behavior is _slightly different_. 266 attempted to make use of the new generic interface we generate with Stencil. These interfaces are not exported by default, so we cannot rely on trying to import them from the source project (unless we added some sort of flag + docs for developers to use it). I opted against this for now. We can re-evaluate adding support for that type when v3 of Stencil ships. 

By not using that generic interface, the `.target` will not be correctly typed to the web component's `HTMLElement` (when using event bindings in Angular). 

Dev-build: `0.6.1-dev.11667512184.13935f80`

### DIFF
--- a/packages/angular-output-target/angular-component-lib/utils.ts
+++ b/packages/angular-output-target/angular-component-lib/utils.ts
@@ -36,6 +36,28 @@ export const defineCustomElement = (tagName: string, customElement: any) => {
   }
 };
 
+/**
+ * The Angular property name that contains the object of metadata properties
+ * for the component added by the Angular compiler.
+ */
+const NG_COMP_DEF = 'ɵcmp';
+
+export const clearAngularOutputBindings = (cls: any) => {
+  if (typeof cls === 'function' && cls !== null) {
+    if (cls.prototype.constructor) {
+      const instance = cls.prototype.constructor;
+      if (instance[NG_COMP_DEF]) {
+        /**
+         * With the output targets generating @Output() proxies, we need to
+         * clear the metadata (ɵcmp.outputs) so that Angular does not add its own event listener
+         * and cause duplicate event emissions for the web component events.
+         */
+        instance[NG_COMP_DEF].outputs = {};
+      }
+    }
+  }
+};
+
 // tslint:disable-next-line: only-arrow-functions
 export function ProxyCmp(opts: { defineCustomElementFn?: () => void; inputs?: any; methods?: any }) {
   const decorator = function (cls: any) {
@@ -44,6 +66,8 @@ export function ProxyCmp(opts: { defineCustomElementFn?: () => void; inputs?: an
     if (defineCustomElementFn !== undefined) {
       defineCustomElementFn();
     }
+
+    clearAngularOutputBindings(cls);
 
     if (inputs) {
       proxyInputs(cls, inputs);
--- a/packages/angular-output-target/__tests__/generate-angular-component.spec.ts
+++ b/packages/angular-output-target/__tests__/generate-angular-component.spec.ts
@@ -1,3 +1,4 @@
+import { ComponentCompilerEvent } from '@stencil/core/internal';
 import { createComponentTypeDefinition, createAngularComponentDefinition } from '../src/generate-angular-component';
 
 describe('createAngularComponentDefinition()', () => {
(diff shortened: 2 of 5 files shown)


sean-perkins:
I don't have the time this sprint, but I have a few ideas to remove this "workaround" with clearing the decorated metadata; which would be my ideal state of this change.

I _think_ we could do away with the `proxyOutputs` and instead map the `fromEvent` directly to the generated `@Output` in place of `EventEmitter`. An `EventEmitter` is just an extended RxJS Subject in Angular.

e.g.:
```ts
@Output() myEvent = fromEvent(this.el, 'my-event');
```

I'm just unsure if I can access the element reference at the type of class construction - which is where the additional time would be needed to diagnose/test.

packages/angular-output-target/angular-component-lib/utils.ts
```diff
@@ -36,6 +36,28 @@ export const defineCustomElement = (tagName: string, customElement: any) => {
   }
 };
 
+/**
+ * The Angular property name that contains the object of metadata properties
+ * for the component added by the Angular compiler.
+ */
+const NG_COMP_DEF = 'ɵcmp';
```
liamdebeasi, line 43: Do we have any concerns about this potentially changing across major Angular releases?

packages/angular-output-target/angular-component-lib/utils.ts
```diff
@@ -36,6 +36,28 @@ export const defineCustomElement = (tagName: string, customElement: any) => {
...
+
+export const clearAngularOutputBindings = (cls: any) => {
```
liamdebeasi, line 45: Can we add types here?

estankovic:
Hi, I would like to ask, what is the status of this PR?

packages/angular-output-target/angular-component-lib/utils.ts
sean-perkins: If the API does change, the patch surrounding this would more or less not be applied, which would default to the existing behavior developers experience today. `ɵcmp` has been used within Angular for several major releases (since v11-12 era). If it does change, likely this fix would not be needed, since it is tied into how Angular stores metadata for the Angular Language Service + event/property bindings.

packages/angular-output-target/angular-component-lib/utils.ts
sean-perkins: The only type addition that I am familiar with that I could see applying here is something like:

```ts
type Clazz = new (...args: any[]) => unknown;

export const clearAngularOutputBindings = (cls: Clazz) => {
  ..
}
```

but this would only type `cls.prototype`. After that the inferred types are `any`, so it doesn't provide type safety where necessary. 

Also the calling implementation uses `any`: https://github.com/ionic-team/stencil-ds-output-targets/blob/34f652daf77703519d0036b4da7028756edab7b8/packages/angular-output-target/angular-component-lib/utils.ts#L63-L70 which doesn't really benefit this change.

ofirrifo:
Hi, I would like to ask, what is the status of this PR?

liamdebeasi (approved):

eliaharris:
@sean-perkins Is this PR ready to go? I agree with @tanner-reits that the output looks great, and I'm not seeing any breaks when testing.

tfrijsewijk:
I tried this with our Stencil library and it works great 👏 

image

Although there's still no JSDoc for Inputs

image

It's still a big step forward! 🥇

alexandertrefz:
Is there anything blocking this?

sean-perkins:
We have discussed this change with the Angular team and have been advised not to make use of this internal API to workaround the issue. 

This issue has moved into our backlog and won't be rescheduled until Ionic Framework has released v7, so that we have enough time to investigate and test alternative solutions. 

I am going to close out this PR and will open a new PR when the updated approach is available for review.

FabioGimmillaro:
I commented in [#303](https://github.com/ionic-team/stencil-ds-output-targets/issues/303)

https://github.com/ionic-team/stencil-ds-output-targets/issues/303#issuecomment-1541522469

Could this be a viable solution for this problem?

tfrijsewijk:
I have re-applied the changes to version 0.7.1 as I was still unable to get typings for events. Perhaps it's me, but I don't understand how anyone is working with Angular bindings without this change 🤷

FabioGimmillaro:
Hi @sean-perkins 

Ionic v7 is already out. Is there already a new issue for this bug?

tfrijsewijk:
Any news? I'm still wondering how anyone is working with Stencil generated Angular bindings. Is everyone ignoring the type errors?

PR #318, opened 2023-01-05
chore(deps): update semver ranges for @stencil/core package

alicewriteswrongs:
## Pull request checklist

- [ ] Build (`npm run build`) was run locally for affected output targets
- [ ] Prettier (`npm run prettier`) was run locally and passed

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL:

## What is the new behavior?

This change updates the semver ranges for the `@stencil/core` peer dependency for various framework wrapper packages. This should let us use them with stencil v3.

- [ ] Yes
- [x] No

## testing

I think this all works, but to be honest I am not the biggest expert on this repo.

Here's what I did to check that it's working.

### building output targets

First we need to build and package test versions of the framework wrapper packages which have the new version.

This will bootstrap the repo, run build in all the packages, and then run `npm pack` and `npm link` in each of them (I'm not super familiar with lerna so very probably there is a better way to do this!):

```sh
npm run bootstrap
npm run build
npx lerna exec npm pack
npx lerna exec npm link
```

After that we should have a `.tgz` file for each built and set up so that we can install it on the framework side.

### install in framework

Ok so basically we'll need to install the packages we just built in framework and install Stencil v3.

First we need to link the framework wrapper packages we just built. Clone the framework repo and in the root of it do the following:

```sh
cd core
npm i
npm link @stencil/angular-output-target --save
npm link @stencil/react-output-target --save
npm link @stencil/vue-output-target --save
```

You may not need the `--save` flag - I ran into a confusing interaction between `npm`, `lerna`, and `volta` and this was the way that I got it to work 🤷‍♀️ 

Once you have the framework wrapper packages installed then you'll want to install the v3 beta of Stencil, like so:

```sh
npm i --save @stencil/core@3.0.0-beta.0 --force
```

You should still see some peer dependency errors, but they should only be from `@stencil/sass` and not from `@stencil/angular-output-target` and company.

Ok finally we can run a build in Framework:

```
npm run build
```

### build and test react, angular, and vue packages in framework

Now we need to test that we can build and test the framework-specific packages for the Ionic framework. We can lean on Lerna for a bunch of this.

First I had to do this to get package dependencies to install correctly:

```sh
npx lerna exec "npm install --legacy-peer-deps"
```

then we can build all the packages by doing

```sh
npx lerna run sync
npx lerna run build --ignore @ionic/docs
```

🚨⚠️**NOTE**⚠️🚨 at present the build does _not_ work because of a known issue with circular type exports. However, repeating the above steps with `@stencil/core@2.18.0` installed does work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   },
   "devDependencies": {
     "@ionic/prettier-config": "^2.0.0",
-    "@stencil/core": "^2.9.0",
+    "@stencil/core": "^2.9.0 || ^3.0.0-beta.0",
     "@types/jest": "^26.0.9",
     "jest": "^26.3.0",
     "lerna": "^4.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,7 +7,7 @@
       "name": "root",
       "devDependencies": {
         "@ionic/prettier-config": "^2.0.0",
-        "@stencil/core": "^2.9.0",
+        "@stencil/core": "^2.9.0 || ^3.0.0-beta.0",
         "@types/jest": "^26.0.9",
         "jest": "^26.3.0",
         "lerna": "^4.0.0",
--- a/packages/angular-output-target/package-lock.json
+++ b/packages/angular-output-target/package-lock.json
@@ -13,7 +13,21 @@
         "@angular/forms": "8.2.14"
       },
       "peerDependencies": {
-        "@stencil/core": "^2.9.0"
+        "@stencil/core": "^2.9.0 || ^3.0.0-beta.0"
+      }
+    },
+    "node_modules/@angular/common": {
+      "version": "8.2.14",
+      "resolved": "https://registry.npmjs.org/@angular/common/-/common-8.2.14.tgz",
+      "integrity": "sha512-Qmt+aX2quUW54kaNT7QH7WGXnFxr/cC2C6sf5SW5SdkZfDQSiz8IaItvieZfXVQUbBOQKFRJ7TlSkt0jI/yjvw==",
+      "dev": true,
+      "peer": true,
+      "dependencies": {
+        "tslib": "^1.9.0"
+      },
+      "peerDependencies": {
+        "@angular/core": "8.2.14",
+        "rxjs": "^6.4.0"
       }
     },
     "node_modules/@angular/core": {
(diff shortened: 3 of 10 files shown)


package.json
```diff
@@ -15,7 +15,7 @@
   },
   "devDependencies": {
     "@ionic/prettier-config": "^2.0.0",
-    "@stencil/core": "^2.9.0",
```
rwaskiewicz, line 18: I don't _think_ NPM knows how to bump packages with text in their version string. e.g. I'm not sure if it knows that `v3.0.0-rc.0` (the next type of release we'll do) is a higher version than `v3.0.0-beta.0`. But maybe I'm wrong there - do you know if that's the case?

Ideally, I don't want to have to put up a PR when we go from beta->rc->ga, mostly so we don't have as tight of a coordination schedule when we're rolling v3.0.0 out. In the event that NPM doesn't know that rc comes after beta (and that `3.0.0` comes after rc), I'm not sure what the best course of action is here. IIRC npm can get finicky here. Maybe using `v3-next` in our accepted versions string in addition to `^3.0.0`? I'm not 100% sure npm will accept that without doing something like:
1. Installing v3-next with `npm i @stencil/core@v3-next`
2. Manually edit the dependencies
3. Installing the latest Stencil Core (on v2) with `npm i @stencil/core@latest`)

rwaskiewicz, line 18: Should we update the `package-lock.json` files here (and in the other files) as well? `devDependencies` and `peerDependencies` still get written to the manifest when running `npm i`. I think this might get a little tricky with respect to my other comment on this line though 😢

rwaskiewicz (changes requested):

package.json
alicewriteswrongs: Yeah I'm not 100% sure on some of these range matching bits, but I went ahead with this range after checking here: https://jubianchi.github.io/semver-check/#/^3.0.0-beta.0/3.0.0

and doing some reading of the `semver` package documentation, here: https://github.com/npm/node-semver

I _believe_ that, for instance, `3.0.0` or `3.0.1` will be accepted as meeting `^3.0.0-beta.0` (as will, _I think_, `3.0.0-beta.1`, `3.0.0-rc.0`, etc).

alicewriteswrongs: ah yeah that's probably a good idea 😅 

I think I missed it just because I was `lerna`-ing and I thought it would do that automatically!

alicewriteswrongs:
For the `package-lock.json` I'm not sure, I just ran an `npm install` in each package and then committed what changed

rwaskiewicz:
Ah, I suspect that these haven't been updated in that case :-(

Well, I suppose we need to update at some point, might as well do it now 🚢

rwaskiewicz (approved):

PR #349, opened 2023-06-02
chore(npm): manually bumping the package versions

JessicaSachs:
PRs like this are necessary until we have a bot committing back onto main after a successful production release.

### DIFF
--- a/packages/angular-output-target/package-lock.json
+++ b/packages/angular-output-target/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@stencil/angular-output-target",
-  "version": "0.7.0",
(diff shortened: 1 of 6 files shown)


rwaskiewicz (changes requested):
Can we bump the `package-lock.json` file versions as well? IIRC there are two places per file that need to get updated, but running `npm version --allow-same-version <PKG_VERSION>` in each project should  do the trick (aside from manually updating them)

rwaskiewicz (approved):

sean-perkins (approved):

tanner-reits (approved):

PR #366, opened 2023-07-21
feat(angular-output-target): generate standalone components

sean-perkins:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue URL: Internal

N/A

## What is the new behavior?

- Angular output target can generate standalone Angular components when using `outputType: "scam"` and using the custom elements build output from Stencil (specifying a `customElementsDir`).

- [ ] Yes
- [x] No

### DIFF
--- a/packages/angular-output-target/src/generate-angular-component.ts
+++ b/packages/angular-output-target/src/generate-angular-component.ts
@@ -10,6 +10,7 @@ import { createComponentEventTypeImports, dashToPascalCase, formatToQuotedList }
  * @param outputs The outputs/events of the Stencil component. (e.g. ['myOutput']).
  * @param methods The methods of the Stencil component. (e.g. ['myMethod']).
  * @param includeImportCustomElements Whether to define the component as a custom element.
+ * @param standalone Whether to define the component as a standalone component.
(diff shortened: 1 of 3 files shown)


liamdebeasi (approved):

packages/angular-output-target/src/generate-angular-component.ts
```diff
@@ -48,6 +50,12 @@ export const createAngularComponentDefinition = (
     proxyCmpOptions.push(`\n  methods: [${formattedMethods}]`);
   }
 
+  let standaloneOption = '';
+
+  if (standalone && includeImportCustomElements) {
+    standaloneOption = `\n  standalone: true`;
```
sean-perkins, line 56: We take this approach because `standalone` is a "newer" feature that isn't a valid property of the `Component` decorator in older versions of Angular. By conditionally applying both the property and the value we accomplish a few things:
1. The Angular output target is compatible with a greater range of Angular versions.
2. We reduce the generated boilerplate for projects not using standalone components.

rwaskiewicz (approved):

PR #422, opened 2024-02-29
Update GitHub, author and package information for fork

sdvg:
package-lock files are removed for the moment because they would be in an impossible state until new packages are published.


sdvg:
Sorry for the noise, wrong PR target.

PR #425, opened 2024-03-25
fix(vue): v-model does not update when other events bubble up

### LINKED ISSUE #424: bug: Vue IonAccordionGroup takes value from underlying component
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/ionic-framework/blob/main/.github/CONTRIBUTING.md#creating-an-issue).
- [X] I agree to follow the [Code of Conduct](https://ionicframework.com/code-of-conduct).
- [X] I have searched for [existing issues](https://github.com/ionic-team/ionic-framework/issues) that already report this problem, without success.

### Ionic Framework Version

v7.x

### Current Behavior

It appears that the ion-accordion-group v-model takes the value from an underlying component (ion-searchbar in this case) when the underlying component changes model value.

Seems this is the same issue as https://github.com/ionic-team/ionic-framework/issues/28288

### Expected Behavior

Model value of ion-accordion-group should not be affected by a changing model value of an underlying component

### Steps to Reproduce

```html
<template>
  <ion-accordion-group v-model="accordion">
    <ion-accordion value="one">
      <ion-item slot="header">
        <ion-label>One</ion-label>
      </ion-item>
      
      <div slot="content">        
        <ion-searchbar v-model="search"></ion-searchbar>
      </div>
    </ion-accordion>

    <ion-accordion value="two">
      <ion-item slot="header">
        <ion-label>Two</ion-label>
      </ion-item>

      <div slot="content">        
        <h1>test</h1>
      </div>
    </ion-accordion>
  </ion-accordion-group>

  <br>
  <span>
    Accordion value: {{ accordion }}
  </span>
</template>

<script lang="ts">
  import { markRaw } from 'vue';
  import { IonAccordionGroup, IonAccordion, IonSearchbar } from '@ionic/vue';

  export default {
    components: { IonAccordionGroup, IonAccordion, IonSearchbar },
    data() {
      return {
        accordion: null,
        search: ''
      };
    },
  };
</script>
```

1. Open accordion `One`. Text below the accordion will change to `one`.
2. Type in any text in the searchbar
3. Click outside the accordion anywhere, like below it. 
4. Accordion closes, because the `accordion` value changed to the text that was typed into the searchbar.

### Code Reproduction URL

https://stackblitz.com/edit/hhmqhk-nfvvdf?file=src%2Fcomponents%2FExample.vue

### Ionic Info

Ionic:

   Ionic CLI       : 6.20.5 (C:\Users\Chris\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework : @ionic/vue 7.8.1

Capacitor:

   Capacitor CLI      : 5.5.0
   @capacitor/android : 5.5.0
   @capacitor/core    : 5.5.0
   @capacitor/ios     : 5.5.0

Utility:

   cordova-res : not installed globally
   native-run  : 1.7.3

System:

   NodeJS : v20.11.1 (C:\Program Files\nodejs\node.exe)
   npm    : 10.2.4
   OS     : Windows 10

### Additional Information

_No response_

liamdebeasi:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: resolves #424

## What is the new behavior?

- The Vue wrappers only update the v-model on an element when the event originated from that element and not from any descendant elements.

- [ ] Yes
- [x] No

## Other information

Dev build: `0.0.1-dev.11711372539.10de27e3`

Verified this issue is fixed using an Ionic Framework build in the sample app from the linked issue.

### DIFF
--- a/packages/vue-output-target/vue-component-lib/utils.ts
+++ b/packages/vue-output-target/vue-component-lib/utils.ts
@@ -91,8 +91,17 @@ export const defineContainer = <Props, VModelType = string | number | boolean>(
         const eventsNames = Array.isArray(modelUpdateEvent) ? modelUpdateEvent : [modelUpdateEvent];
         eventsNames.forEach((eventName: string) => {
           el.addEventListener(eventName.toLowerCase(), (e: Event) => {
-            modelPropValue = (e?.target as any)[modelProp];
-            emit(UPDATE_VALUE_EVENT, modelPropValue);
+            /**
+             * Only update the v-model binding if the event's target is the element we are
+             * listening on. For example, Component A could emit ionChange, but it could also
+             * have a descendant Component B that also emits ionChange. We only want to update
+             * the v-model for Component A when ionChange originates from that element and not
+             * when ionChange bubbles up from Component B.
+             */
+            if (e.target.tagName === el.tagName) {
+              modelPropValue = (e?.target as any)[modelProp];
+              emit(UPDATE_VALUE_EVENT, modelPropValue);
+            }
           });
         });
       },


packages/vue-output-target/vue-component-lib/utils.ts
```diff
@@ -91,8 +91,17 @@ export const defineContainer = <Props, VModelType = string | number | boolean>(
...
-            modelPropValue = (e?.target as any)[modelProp];
-            emit(UPDATE_VALUE_EVENT, modelPropValue);
+            /**
+             * Only update the v-model binding if the event's target is the element we are
+             * listening on. For example, Component A could emit ionChange, but it could also
+             * have a descendant Component B that also emits ionChange. We only want to update
+             * the v-model for Component A when ionChange originates from that element and not
+             * when ionChange bubbles up from Component B.
+             */
+            if (e.target.tagName === el.tagName) {
```
rwaskiewicz, line 101: Is it guaranteed that at least one of these two values can never be `undefined`? My only thought here is running into a case where we end up doing:
```
if (undefined === undefined) {
```

rwaskiewicz (approved):

packages/vue-output-target/vue-component-lib/utils.ts
liamdebeasi: Good question. `Element.tagName` is [always defined](https://www.typescriptlang.org/play?#code/MYewdgzgLgBApgGxgXhgExMArgWzmKAOmACc4BDKOAUQTjwIAoByNASwDdmBKAbgCh+iQlHIBzAHLk8QA). 

If `e.target` is defined then it should be an `Element` (which means `tagName` is defined). However, `EventTarget` can [potentially be `null`](https://www.typescriptlang.org/play?#code/MYewdgzgLgBApgNxgXhmOB3GBhArtEAWwFEE4woAKAcgDMQRqBKAbgCg3EA6KAQwCcA5nChA). My understanding is that `target` can be null when an event is created but not dispatched on an element. You can test this by logging `new CustomEvent('foo')`. In this case, the event is dispatched on an element, so `target` should be non-null.

tanner-reits (approved):

sean-perkins (approved):

PR #445, opened 2024-07-03
feat(react): Next.js Support

christian-bromann:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The current React output target can only be used on the client side. There is no support for server side rendered content, e.g. within a Next.js environment

## What is the new behavior?

This patch enhances the React output target to support server side rendering.

- [ ] Yes
- [x] No

## Other information

This is work in progress.


christian-bromann:
PR with additional documentation can be found in https://github.com/ionic-team/stencil-site/pull/1458

packages/example-project/component-library/stencil.config.ts
```diff
@@ -74,9 +78,16 @@ export const config: Config = {
...
       esmLoaderPath: '../loader',
     },
+    {
+      type: 'dist-hydrate-script',
+      dir: './hydrate',
+    },
     {
       type: 'docs-readme',
     },
+    {
```
sean-perkins, line 88: Does the hydrate script output target need to be defined twice?

tanner-reits (approved):
LGTM. This is a good one to get some people testing before we merge all the way back to main, which I think is the plan anyway

laurencefass:
great news! do you have any further info on what exactly Next JS support looks like. Im currently using a Provider pattern so though I render all Ionic components on the client I am able to load the context through the root layout so benefit from everything Next JS app router has to offer. 

I guess some components will have to be rendered on the client i.e. those with interactivity. I anticipate thats quite a lot of components. 

Is there a discussion on this which can be used as the foundation for docs later on, or at least to manage expectations during testing now?

christian-bromann:
> do you have any further info on what exactly Next JS support looks like. Im currently using a Provider pattern so though I render all Ionic components on the client (with 'use client').

With this patch the react output target will allow you to create a Next.js framework wrapper which enables the usage of Stencil components within Next.js applications including support for SSR, so that you can use Stencil components also on the server side.

> Is there a discussion on this which can be used as the foundation for docs later on, or at least to manage expectations during testing now?

I think this PR thread is the best for posting comments and providing feedback on this development.

@mayerraphael I've just created this dev build: `@stencil/react-output-target@0.0.1-dev.11720819946.1e46a246` that you can use for testing. Note that I am still working on some render issues, e.g. Stencil seems to render scoped components twice for a short period until it reconciles properly. Also there is a hydration error being thrown when using scoped components. I am looking into that now.

christian-bromann:
> It looks like Stencil uses dist/components as default and the Output Target uses /components as default

Good catch, fixed.

> Events that had an underscore prefix got broken inside the generated components.ts file.

I will take a look at that.

> Server render errors

I pushed a potential fix in this new dev version: `0.0.1-dev.11721063914.1cb11145`, mind taking another stab at it? It also includes a fix for the default path.

laurencefass:
I dont really understand whats happening here. As a Next developer just wanting to use Ionic components in my app, my (possibly oversimplistic?) expectation is to know what Ionic components I can render on the server and which are client only. There seems to be a lot more going on here at a lower level of abstraction. Is the outcome of this being able to render (some?) Ionic components on the server?

christian-bromann:
@mayerraphael did you check with the latest dev version? `0.0.1-dev.11721063914.1cb11145` has a patch that resolves all async children. 

> Is the outcome of this being able to render (some?) Ionic components on the server?

Yes.

laurencefass:
```
Is the outcome of this being able to render (some?) Ionic components on the server?

Yes.
```
So...  a finite (documented?) set of layout only Ionic server (isomorphic?) components?

christian-bromann:
First off, thanks a lot @mayerraphael for all the feedback!

> Please put that it is only compatible with the AppRouter into your documentation or i guess you might get some bug tickets :)

Yes, for now we will only support AppRouter as Stencils serialization function is async and I assume it would require larger efforts to support a sync version of that.

> GlobalScripts execution missing

I am not sure if global scripts should be part of the hydration process. It is hard to say as I don't have too much knowledge about what people do in the globalscripts but I believe it is mostly setting up the browser environment which is not needed on the server side. For example there is no need to call `defineCustomElements` when server side rendering the component as the hydrate module already has all components registered.

I personally would not recommend the use of global scripts at all as they can potentially block the app from being rendered and slow down your app load performance.

> But we also have global scripts which ran with `defineCustomElements()` (loader) before.

This way this is meant to be used is: the React output target creates a `component.ts` files which basically wraps all Stencil components into a React component which you can import and use in other React components, e.g. via `<MyComponent />`. You only need to call `defineCustomElements` when you want to use the raw web components, e.g. via `<my-component></my-component>`. The point of the React output target is to make this all simpler and use the React wrapper.

> It's not possible to run the new React Output Target with the old `renderToString` and NextJS PageRouter anymore (see why in my previous comment).

I am not sure how you've used the React Output Target before but I believe it always has been this way that it would generate wrapper files. The current stable version of the React output target doesn't use `renderToString` at all. So I am not quite sure what you mean by this.

> * Allow user to specify if `dist-custom-element` or `loader` (like before, default) is used

You still can do that. In your `stencil.config.ts` you can define `dist`, `dist-custom-element` and the React Output Target and get artifacts for all of them. This will not change with the new version of this package.

> Allow user to specify if `renderToString` is called internally or called by the user on the html (like before, default)
> * This is an opt-out of the new SSR concept. The client component is used on the server, like before.

Note that this new feature is mainly to be used for meta frameworks such as Next or Remix (haven't tested on that one yet). You can still run your own hydration process using `renderToString` manually, nothing has changed here (except that we have shipped updates to Stencil which now properly renders shadow components as declarative shadow DOM.

> * `globalScript` must be run or exported so it can be run by the user (not required if `loader` is used).

Not sure if I understand this issue. Do you have a concrete use case?

> The new concept should either be a `flag` in the `reactOutputTarget options` or a completly independend package.
> The manual way (calling hydrate.renderToString) should still work.

The SSR support for the React Output Target is behind a flag (`hydrateModule`). You are always able to manually render a component using `renderToString` as this is a Stencil feature and has not much to do with the React Output Target.

> Is there a reason why the latest stable react-output-target is 0.5.3 from a year ago and newer dev version are available (like the one we test now) but never made it to stable?

There hasn't been any updates for quite some time. Just recently we have updated the React Output Target which we are still testing internally including this new feature.

christian-bromann:
> How can i run that script before the React components define and load themself on the client?

I would suggest to move this into the [`connectedCallback`](https://stenciljs.com/docs/component-lifecycle#connectedcallback) lifecycle method. Can you elaborate what you want to do in the global script? In general I would not recommend to use the global script as it blocks rendering the application as a whole. From the user perspective it makes much more sense to handle e.g. data loading within the component and provide some sort of loading state to have a visual feedback to users. However I am not sure what your use case is. Furthermore I don't think global scripts ever worked in combination with framework wrappers like the React output target, let me verify this assumption though.

> generated **4MB** of payload. All components are included in the page chunks' generated production build. The whole of `dist-custom-elements`

Let me take a look to see if we can tree shake the code better here. Thanks for pointing this out!

christian-bromann:
> How can i run that script before the React components define and load themself on the client?

I was able to confirm with a @sean-perkins that global scripts were never supported when using Stencil through these framework wrappers. Again, I would love to hear more details about the use case but I generally would recommend to run that logic as part of the application loading process within React or in the `connectedCallback` lifecycle as part of a single component.

christian-bromann:
@mayerraphael can you try with
```
@stencil/core@4.19.2-dev.1722290246.61e8645
@stencil/react-output-target@0.0.1-dev.11722291371.147e648d
```

christian-bromann:
@mayerraphael yes, I recognised this last week and actually added a check into the output target to validate that but it didn't recognised that `true` is the default value. Let me know if you discover anything weird.

christian-bromann:
@mayerraphael I have a new PR open with some improvements on the implementation, mind taking a look at providing feedback if provided dev build (see PR description) works better?

packages/react-output-target/src/react-output-target/create-component-wrappers.ts
```diff
@@ -9,18 +11,18 @@ export const createComponentWrappers = async ({
...
   excludeComponents,
   project,
+  hydrateModule,
 }: {
   stencilPackageName: string;
   components: ComponentCompilerMeta[];
   customElementsDir: string;
   outDir: string;
   esModules?: boolean;
-  experimentalUseClient?: boolean;
```
sean-perkins, line 21: Will older environments be able to parse (or skip) the `use client;` directive annotation in JS files?

sean-perkins (approved):
Great work! I don't see anything that is blocking. 

I have one question on the compatibility with environments.

packages/react-output-target/src/react-output-target/create-component-wrappers.ts
sean-perkins: Ahh I see, the `use client;` was moved to only be generated as part of the Next.js component wrappers. The config option was replaced with the hydrate module. 

This makes sense - I don't have concerns 👍

packages/react-output-target/README.md
```diff
@@ -85,5 +85,7 @@ That's it! You can now import and use your Stencil components as React component
...
-| `experimentalUseClient` | If `true`, the generated output target will include the [`use client;`](https://react.dev/reference/react/use-client) directive.                                                                                                                                               |
+
 | `customElementsDir`     | The directory where the custom elements are saved. This value is automatically detected from the Stencil configuration file for the `dist-custom-elements` output target. If you are working in an environment that uses absolute paths, consider setting this value manually. |
+
+| `hydrateModule`         | For server side rendering support, provide the package for importing the [Stencil Hydrate module](https://stenciljs.com/docs/hydrate-app#hydrate-app), e.g. `my-package/hydrate`. This will generate two files: a `component.server.ts` which defines all component wrappers and a `components.ts` that re-exports these components for importing in your application. |
```
smorrisods, line 91: Sorry to be late to the party but these additions to the table aren't included in it due to the extra new lines between them and the existing table.

image

christian-bromann: @smorrisods thanks for pointing this out, mind raising a PR?

smorrisods: I'd be happy to!

smorrisods: Upon further investigation it looks like it has already been fixed in commit 75c6cce4cc927404335f5d400678200008ed4230! No further PRs needed 🙂.

PR #446, opened 2024-08-01
fix(vue): pass router-link value to href to properly render clickable elements

brandyscarney:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Ionic Framework Vue components using `router-link` do not apply an `href` property which causes components to render `div` or `button` elements when they should render an `a`. 

## What is the new behavior?
Check if `router-link` and `navManager` are defined so this only applies to Ionic Framework components. If both are defined then add the `href` property to the element with the value of `router-link`.

- [ ] Yes
- [x] No


sean-perkins (approved):
LGTM, I watched you make the changes and verified them.

PR #447, opened 2024-08-03
Next.js SSR Hydration with Children

christian-bromann:
This PR is an experimental enhancement to #445 which considers the light DOM when serializing Stencil components. We've experienced that a Stencil component may consider the light DOM for certain rendering conditions, e.g. add style properties or class names to a host component if the light DOM contains a certain slot.

You can use this dev build `@stencil/react-output-target@0.0.1-dev.11722646072.12d0ef0b` for testing purposes. Any feedback is much appreciated!

### DIFF
--- a/packages/example-project/next-app/src/app/Input/Input.tsx
+++ b/packages/example-project/next-app/src/app/Input/Input.tsx
@@ -12,7 +12,10 @@ function Input() {
       <MyInput
         onMyInput={(ev) => setInputEvent(`${ev.target.value}`)}
         onMyChange={(ev) => setChangeEvent(`${ev.detail.value}`)}
-      />
+      >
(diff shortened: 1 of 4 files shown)


packages/example-project/next-app/src/app/Input/Input.tsx
```diff
@@ -12,7 +12,10 @@ function Input() {
       <MyInput
         onMyInput={(ev) => setInputEvent(`${ev.target.value}`)}
         onMyChange={(ev) => setChangeEvent(`${ev.detail.value}`)}
-      />
+      >
+        {/* the following space makes the hydration error go away */}
+        {' '}
```
tanner-reits, line 17: This a limitation we need to call out anywhere?

PR #437, opened 2024-05-17
feat: add value accessors to the directives array

AndreasIllg-COS:
The value accessor directives also need to be declared in the module file. This is achieved here by adding them to the directives array file, since it is assumed that the module declares and exports all elements of this array.

## Pull request checklist
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally for affected output targets
- [x] Tests (`npm test`) were run locally and passed
- [x] Prettier (`npm run prettier`) was run locally and passed

## Pull request type
Please check the type of change your PR introduces:

- [ ] Bugfix
- [X] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?
The value accessor directives have to be added to the angular module manually.

Issue URL: https://github.com/ionic-team/stencil-ds-output-targets/issues/436

## What is the new behavior?
The value acessor directives are included in the directivesArrayFile and thereby declared and exported by a module file as described in the angular integration guide (https://stenciljs.com/docs/angular#adding-the-angular-output-target)

## Does this introduce a breaking change?

- [X] Yes
- [ ] No

The value accessors need to be removed from the public-api exports, since they are not copied any more. 
The only export necessary hereafter is 
```
export * from './lib/stencil-generated/value-accessors';
```
This line should added to the angular integration guide (https://stenciljs.com/docs/angular#adding-the-angular-output-target)

PR #459, opened 2024-08-22
feat(angular): Standalone Value Accessor for Angular OutputType

Samg983:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [X] Code style update (formatting, renaming)
- [ ] Build related changes

There is no support for standalone value accessors for Angular.

Issue URL: https://github.com/ionic-team/stencil-ds-output-targets/issues/460

## What is the new behavior?

- When it's outputType is standalone or scam, the value accessor will also be standalone

- [ ] Yes
- [X] No


packages/angular-output-target/src/generate-value-accessors.ts
```diff
@@ -54,30 +55,34 @@ export default async function generateValueAccessors(
...
       VALUE_ACCESSOR_TARGETATTR,
       listItem[1]
     )
   );
-
   return srcFileContents
     .replace(VALUE_ACCESSOR_SELECTORS, valueAccessor.elementSelectors.join(', '))
-    .replace(VALUE_ACCESSOR_EVENTTARGETS, hostContents.join(`,${EOL}`));
+    .replace(VALUE_ACCESSOR_EVENTTARGETS, hostContents.join(`,${EOL}`))
+    .replace(VALUE_ACCESSOR_STANDALONE, outputType && outputType !== OutputTypes.Component ? ',standalone: true' : '');
```
tanner-reits, line 80: I think we need to update this check so it _only_ applies standalone for `OutputTypes.Standalone`, otherwise it'll set standalone to `true` if they supply `OutputTypes.Scam`, which I don't think is appropriate.

```suggestion
    .replace(VALUE_ACCESSOR_STANDALONE, outputType && outputType === OutputTypes.Standalone ? ',standalone: true' : '');
```

Would be good to add a test for this as well

packages/angular-output-target/__tests__/generate-value-accessors.spec.ts
```diff
@@ -41,6 +41,47 @@ export class TextValueAccessor extends ValueAccessor {
     super(el);
   }
 }`;
-    expect(finalText.trim()).toEqual(exptectedOutput.trim().replace(/\n/g, EOL));
+    console.log(expectedOutput.trim().replace(/\n/g, EOL));
+    console.log(finalText.trim());
```
tanner-reits, line 45: Can we remove these logs, please 😄

packages/angular-output-target/resources/control-value-accessors/boolean-value-accessor.ts
```diff
@@ -15,7 +15,7 @@ import { ValueAccessor } from './value-accessor';
       useExisting: BooleanValueAccessor,
       multi: true
     }
-  ]
+  ]<VALUE_ACCESSOR_STANDALONE>
```
tanner-reits, line 18: For readability, it might be clearer to update this so that we only replace the boolean value for the `standalone` option. So something like:

```suggestion
  ],
  standalone: <VALUE_ACCESSOR_STANDALONE>
```

And then update the replacement logic when generating the accessor.

tanner-reits (changes requested):
Thanks for the PR! Left a couple thoughts. Let me know if you have any questions, or disagree with anything!

packages/angular-output-target/src/generate-value-accessors.ts
Samg983: Wll do!

packages/angular-output-target/__tests__/generate-value-accessors.spec.ts
Samg983: Woops, rookie mistake 🤣

packages/angular-output-target/resources/control-value-accessors/boolean-value-accessor.ts
Samg983: I wouldn't apply this suggestion. I've seen there's a dependency on Angular 8. This standalone feature wasn't implemented in Angular 8 hence it could throw errors. So for backwards compatibility I would keep it as is. What do you think?

tanner-reits: Ah, good call! Makes sense!

tanner-reits (approved):
LGTM! Thanks for your contribution!

christian-bromann (approved):
LGTM 👍

christian-bromann:
I plan to release a version of this today.

christian-bromann:
@Samg983 does this implement what has been suggested in #437?

Samg983:
> @Samg983 does this implement what has been suggested in #437?

It's also an improvement but to me it's not the same. In #437 it's still in a module rather than standalone.

christian-bromann:
This was released in `@stencil/angular-output-target@0.9.0` 🎉

PR #468, opened 2024-08-28
feat(vue): support SSR

christian-bromann:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Currently the Vue Output Target would not support SSR.

## What is the new behavior?

This patch adds support for it. Furthermore:

- it adds a Nuxt.js example application for testing purposes
- it requires users to have `@stencil/vue-output-target` as dependency to access runtime files (similar to React)
- it adds a `hydrateModule` property to the output target that allows users to enable SSR

- [x] Yes
- [ ] No


packages/example-project/component-library/stencil.config.ts
```diff
@@ -66,8 +66,12 @@ export const config: Config = {
       hydrateModule: 'component-library/hydrate'
     }),
     vueOutputTarget({
```
tanner-reits, line 68: Should we have two Vue output target configs like we do for React?

tanner-reits (approved):
There's a lot here, but looks like it works as expected. I tested out the Nuxt app and everything was rendering (some of the components are goofy, but that's just the components, not the SSR implementation). I didn't double-check a client-side Vue app.

christian-bromann:
> I didn't double-check a client-side Vue app.

FWIW I raised PR adding an example app using Vite + Vue, see https://github.com/ionic-team/stencil-ds-output-targets/pull/499. For that I've added an e2e test for this. For all these environments we now run a WebdriverIO tests that verifies that Stencil is properly able to hydrate the component, e.g. interacting with an input component with SSR support or without. While these tests are very rudimental, they are a start a least 😉

packages/example-project/component-library/stencil.config.ts
christian-bromann: React and Vue projects are setup differently here. We have two output targets for React because we directly compile the components into their represented projects. With Vue we have setup an example project that just re-exports the components, and potentially more Vue related modules. That example project is imported by Nuxt and the Vite project. A bit different setup but it is actually how we see projects being organised.

packages/example-project/nuxt-app/public/robots.txt
thetaPC, line 1: Is this file needed?

packages/vue-output-target/src/generate-vue-component.ts
```diff
@@ -1,34 +1,57 @@
...
+    ${Object.entries(propMap)
+      .map(([key, [type, attr]]) => (attr ? `'${key}': [${type}, "${attr}"]` : `'${key}': [${type}]`))
+      .join(',\n    ')}
+  }
+})`
+      : '';
+
+    // tagName: string;
+    // hydrateModule: Promise<{ renderToString: RenderToString }>;
+    // props?: Record<string, any>;
```
thetaPC, line 51: These should be removed if they aren't going to be used.

packages/example-project/nuxt-app/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+# pnpm
+pnpm run dev
+
+# yarn
+yarn dev
+
+# bun
+bun run dev
+```
```
thetaPC, line 39: I'm getting the following error when trying to run it locally: `Missing script: "dev"`. These commands need to be changed to use `start` instead of `dev`.

packages/example-project/nuxt-app/public/robots.txt
christian-bromann: No, it was part of the basic Nuxt setup. I can remove it.

packages/example-project/nuxt-app/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+# pnpm
+pnpm run preview
+
+# yarn
+yarn preview
+
+# bun
+bun run preview
+```
```
thetaPC, line 73: I'm getting this error when trying to visit the preview:
```
Listening on http://[::]:3000
[nuxt] [request error] [unhandled] [500] Cannot find module '/Users/maria/Desktop/projects/stencil-ds-output-targets/packages/example-project/nuxt-app/.output/server/node_modules/vue/server-renderer/index.mjs'
```

thetaPC (approved):
LGTM, added some comments but they're not blockers

PR #484, opened 2024-09-26
fix(react): use property mapping from component meta

christian-bromann:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When server side rendering SSR components, we currently de-camelize the property name which is a wrong assumption, given Stencil user give their properties arbitrary names, e.g. when the `attribute` property is used:

```ts
@Prop({ attribute: 'multi-line', reflect: true, mutable: true }) isMultiLine: boolean = false;
```

## What is the new behavior?

The patch makes the output target pass along the mapping for attributes that Stencil then uses later when it hydrates the component.

- [ ] Yes
- [x] No

## Other information

n/a

### DIFF
--- a/packages/react-output-target/package.json
+++ b/packages/react-output-target/package.json
@@ -16,7 +16,7 @@
     "prepublishOnly": "pnpm run build",
     "build": "vite build && pnpm run build:dts",
     "build:dts": "tsc -p tsconfig.json",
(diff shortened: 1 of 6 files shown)


tanner-reits (approved):
LGTM

Issue #470, opened 2024-08-29
[React] bug: SSR output target duplicates the component's internal shadow DOM layout


dgonzalezr:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil-ds-output-targets/issues) that already report this problem, without success.

### Stencil Version

4.20.0

### Stencil Framework Output Target

React

### Stencil Framework Output Target Version

0.7.1

### Current Behavior

When using the react output target with SSR inside a Nextjs application, the internal shadow DOM container of the component is rendered twice.

![CleanShot 2024-08-29 at 21 48 56](https://github.com/user-attachments/assets/0df0f74a-3e27-4967-9eb9-510a5d2d06d9)

### Expected Behavior

The internal shadow DOM base container should be rendered only once, as expected, depending on the component layout.

### Steps to Reproduce

Follow the steps described in the **Issue 3** of the reproduction code below

### Code Reproduction URL

https://github.com/dgonzalezr/nx-stencil

dgonzalezr:
I can see how the same happens in the [next-app demo of the `stencil-ds-output-target` repo](https://github.com/ionic-team/stencil-ds-output-targets/tree/main/packages/example-project/next-app).

![CleanShot 2024-08-29 at 22 50 24](https://github.com/user-attachments/assets/26e2e9d7-5724-4c12-b339-6bfad113eeef)

christian-bromann:
@dgonzalezr thanks for reporting, I am happy to take a look at this.

dwirz:
[...]

PR #490, opened 2024-10-07
fix(#470): add quasi unique id when creating react SSR component 

dwirz:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

ATM a nested web component leads to a duplication of the root component within the shadow-root of the root component.

Issue URL: https://github.com/ionic-team/stencil-ds-output-targets/issues/470

## What is the new behavior?

This pull request adds a quasi unique id to components which are rendered with `renderToString` in `createComponentForServerSideRendering` so that Stencil can correctly hydrate them client-side.

- [ ] Yes
- [x] No

## Other information

I know that this is not the prettiest and sustainable solution, but it solves the problem temporarily and we could use the package in our project until there is a better solution.

### DIFF
--- a/packages/example-project/component-library-react/src/components.ts
+++ b/packages/example-project/component-library-react/src/components.ts
@@ -9,200 +9,143 @@
 
 import type { EventName, StencilReactComponent } from '@stencil/react-output-target/runtime';
 import { createComponent } from '@stencil/react-output-target/runtime';
-import {
-  type CheckboxChangeEventDetail,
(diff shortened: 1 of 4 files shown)


dwirz:
Hey @christian-bromann, thanks for having a look. Also thanks again for your great work, I'm totally with you to solve this in Stencil is the better approach and your solution is much cleaner and definitely more robust! 🎉  Will test it right away 🤓

PR #514, opened 2024-10-24
Added overrides to boolean and number value accessor as they don't wo…

CosmicEvil:
When adding the value accessors for boolean and number, we are getting errors about the overrides not being there. 
I added them in this commit, as without they won't compile.

### DIFF
--- a/example-project/component-library-angular/src/directives/boolean-value-accessor.ts
+++ b/example-project/component-library-angular/src/directives/boolean-value-accessor.ts
@@ -21,7 +21,7 @@ export class BooleanValueAccessor extends ValueAccessor {
   constructor(el: ElementRef) {
     super(el);
   }
-  writeValue(value: any) {
+  override writeValue(value: any) {
     this.el.nativeElement.checked = this.lastValue = value == null ? false : value;
   }
 }
--- a/example-project/component-library-angular/src/directives/number-value-accessor.ts
+++ b/example-project/component-library-angular/src/directives/number-value-accessor.ts
@@ -21,7 +21,7 @@ export class NumericValueAccessor extends ValueAccessor {
   constructor(el: ElementRef) {
     super(el);
   }
-  registerOnChange(fn: (_: number | null) => void) {
+  override registerOnChange(fn: (_: number | null) => void) {
     super.registerOnChange((value) => {
       fn(value === '' ? null : parseFloat(value));
     });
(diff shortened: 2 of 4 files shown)


christian-bromann:
> I added them in this commit, as without they won't compile.

Can you help me understand where it is not compiling? I don't see any issues within the build. (Please excuse me if this is a naive question as I haven't worked much with the Angular output target yet.)

CosmicEvil:
@christian-bromann Sorry i didn't explain it well enough i think. I can't link to the actual codebase where the issue occurs, as it's a private one, but the gist of it is that i am trying to add those accessors and When adding the Boolean and number one, the error in the issue occurs on build: 

image

I added the overrides to these functions in my "local" version of the npm package, and this fixed the issue, but it is necessary that the fix will also be in the package itself. If you want i can take the time tomorrow to set up a whole new repo that recreates the issue, but this might take a bit of time :)
Thanks in advance!

christian-bromann (changes requested):
It seems like you made the changes in our example projects, not in the actual Angular output target. Can we make the changes in the actual output target, [here](https://github.com/ionic-team/stencil-ds-output-targets/blob/main/packages/angular/resources/control-value-accessors/number-value-accessor.ts#L24), [here](https://github.com/ionic-team/stencil-ds-output-targets/blob/main/packages/angular/resources/control-value-accessors/boolean-value-accessor.ts#L24) and maybe [here](https://github.com/ionic-team/stencil-ds-output-targets/blob/eb8f9d504f621e1437b57dae4dd97873c22e2f67/packages/angular/resources/control-value-accessors/value-accessor.ts#L13) too?

CosmicEvil:
> It seems like you made the changes in our example projects, not in the actual Angular output target. Can we make the changes in the actual output target, [here](https://github.com/ionic-team/stencil-ds-output-targets/blob/main/packages/angular/resources/control-value-accessors/number-value-accessor.ts#L24), [here](https://github.com/ionic-team/stencil-ds-output-targets/blob/main/packages/angular/resources/control-value-accessors/boolean-value-accessor.ts#L24) and maybe [here](https://github.com/ionic-team/stencil-ds-output-targets/blob/eb8f9d504f621e1437b57dae4dd97873c22e2f67/packages/angular/resources/control-value-accessors/value-accessor.ts#L13) too?

Added! The last one was not necessary, as that is the one that is overridden.

christian-bromann (approved):
LGTM 👍

Samg983:
@christian-bromann , any idea when there will be a new release of the output target package?

christian-bromann:
@Samg983 this patch should be already be released. We are working on improving the release pipeline to create better changelogs moving forward.

PR #497, opened 2024-10-14
feat: expose input properties to Angular Language Server

kristilw:
## Pull request checklist

Please check if your PR fulfills the following requirements:

- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build (`npm run build`) was run locally for affected output targets
- [x] Tests (`npm test`) were run locally and passed
- [x] Prettier (`npm run prettier`) was run locally and passed

## Pull request type

Please check the type of change your PR introduces:

- [x] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [x] Documentation content changes
- [ ] Other (please describe):

(I would call this a feature, but it has been reported as a bug)

## What is the current behavior?

Issue URL: https://github.com/ionic-team/stencil/issues/2909

The Angular Language Service is not able to type-check or show jsdocs for input properties. This is because of a limitation of how the Angular Language Service works: https://github.com/angular/angular/issues/48056.

## What is the new behavior?

Angular is able to verifty types on input properties and show jsDocs:
b

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

Feature is hidden behind experimental flag, `inlineProperties`.

I have tested the code on both the example component-library and our own internal design system (which has about 70 components), but there might be some unkown edge cases.

## Other information

Ideally this would be solved by the Angular team, but it has been an open problem for years.

If this MR goes through it is easy to extend the code to support types/jsDocs for output properties as well. The method used to create the correct property declerations are a lot simpler than the current method used to create EventEmitters today, and I believe more robust. The current implementation fails on Array, Promises, functions or when multiple Interfaces shares the same name in the project (such as `Color` in the example component-library). The method used in this MR to create the input proprties does not require any additional import statements either.

PR #533, opened 2024-10-31
fix(angular): 2-way data-binding for inlined Inputs

alexandertrefz:
## Pull request checklist

## Pull request type

- [x] Bugfix

When using the new `inlineProperties` configuration option, properties only update correctly when they are used unbounded(`<my-component prop="value" />`), on the angular side. Bound(`<my-component [prop]="componentProperty" />`) properties are not synced correctly.

## What is the new behavior?

Properties work correctly, no matter if they are used bound or unbounded.

- [x] No

## General Information

PR #497 by @kristilw introduced a long awaited fix for Angular, allowing the Angular Language Server to pick up the types of inputs correctly.

The change works correctly on the Typescript side, as well as at runtime, if the input properties are unbound. However, if they are bound the Angular Compiler does _something_ that prevents the binding to work at all (at runtime), even though the Decorator establishing the getter and setter to pass the value along to the Web Component underneath is run correctly.

This is caused (for some reason) by the property being declared on the class itself, after compilation, like this:

```typescript
export let MyComponent = class MyComponent {
    z;
    el;
    prop; // <-- THIS IS THE PROBLEM
    constructor(c, r, z) {
        this.z = z;
        c.detach();
        this.el = r.nativeElement;
    }
    static ɵfac = function MyComponent_Factory(t) { return new (t || MyComponent)(i0.ɵɵdirectiveInject(i0.ChangeDetectorRef), i0.ɵɵdirectiveInject(i0.ElementRef), i0.ɵɵdirectiveInject(i0.NgZone)); };
    static ɵcmp = /*@__PURE__*/ i0.ɵɵdefineComponent({ type: MyComponent, selectors: [["my-component"]], inputs: { prop: "prop" }, ngContentSelectors: _c0, decls: 1, vars: 0, template: function MyComponent_Template(rf, ctx) { if (rf & 1) {
            i0.ɵɵprojectionDef();
            i0.ɵɵprojection(0);
        } }, encapsulation: 2, changeDetection: 0 });
};
```

However, in order for typescript to create the type in such a way that the (badly implemented) Angular Language Service can pick up the types correctly, we need to declare them on the class.

The solution is to inline the getter and setter for the inputs directly, allowing Angular to pick up the type correctly, and the internal binding to the web component to work correctly as well.

This PR also changes the type of the `el` property, so that it reflects the accurate Web Component type, instead of the generic `HTMLElement`.

### DIFF
--- a/example-project/component-library-angular/src/directives/proxies.ts
+++ b/example-project/component-library-angular/src/directives/proxies.ts
@@ -18,7 +18,7 @@ import { Components } from 'component-library';
   inputs: ['buttonType', 'color', 'disabled', 'download', 'expand', 'fill', 'href', 'mode', 'rel', 'shape', 'size', 'strong', 'target', 'type'],
 })
 export class MyButton {
-  protected el: HTMLElement;
+  protected el: HTMLMyButtonElement;
   constructor(c: ChangeDetectorRef, r: ElementRef, protected z: NgZone) {
     c.detach();
     this.el = r.nativeElement;
@@ -50,7 +50,7 @@ export declare interface MyButton extends Components.MyButton {
   inputs: ['checked', 'color', 'disabled', 'indeterminate', 'mode', 'name', 'value'],
 })
 export class MyCheckbox {
-  protected el: HTMLElement;
+  protected el: HTMLMyCheckboxElement;
   constructor(c: ChangeDetectorRef, r: ElementRef, protected z: NgZone) {
     c.detach();
     this.el = r.nativeElement;
@@ -88,7 +88,7 @@ export declare interface MyCheckbox extends Components.MyCheckbox {
   inputs: ['age', 'favoriteKidName', 'first', 'kidsNames', 'last', 'middle'],
 })
 export class MyComponent {
-  protected el: HTMLElement;
+  protected el: HTMLMyComponentElement;
   constructor(c: ChangeDetectorRef, r: ElementRef, protected z: NgZone) {
     c.detach();
     this.el = r.nativeElement;
@@ -117,7 +117,7 @@ export declare interface MyComponent extends Components.MyComponent {
   inputs: ['accept', 'autocapitalize', 'autocomplete', 'autocorrect', 'autofocus', 'clearInput', 'clearOnEdit', 'color', 'disabled', 'enterkeyhint', 'inputmode', 'max', 'maxlength', 'min', 'minlength', 'mode', 'multiple', 'name', 'pattern', 'placeholder', 'readonly', 'required', 'size', 'spellcheck', 'step', 'type', 'value'],
 })
 export class MyInput {
-  protected el: HTMLElement;
+  protected el: HTMLMyInputElement;
   constructor(c: ChangeDetectorRef, r: ElementRef, protected z: NgZone) {
     c.detach();
     this.el = r.nativeElement;
@@ -160,7 +160,7 @@ export declare interface MyInput extends Components.MyInput {
   inputs: ['animated', 'backdropDismiss', 'component', 'componentProps', 'cssClass', 'event', 'keyboardClose', 'mode', 'showBackdrop', 'translucent'],
 })
 export class MyPopover {
-  protected el: HTMLElement;
+  protected el: HTMLMyPopoverElement;
   constructor(c: ChangeDetectorRef, r: ElementRef, protected z: NgZone) {
     c.detach();
     this.el = r.nativeElement;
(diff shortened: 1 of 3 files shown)


alexandertrefz:
That would explain why you didn't catch this one, I am using `outputType: "component"`, and `type: "dist"`, currently.

dgonzalezr:
Unless I'm missing something, binding property value works for me (with Angular Module):

```html
<bq-progress
  enable-tooltip
  thickness="large"
  [value]="progressValue"
></bq-progress>
```

```ts
export class AppComponent implements AfterViewInit {
  progressValue = 20;

  ngAfterViewInit() {
    this.updateProgress();
  }

  updateProgress() {
    const intervalId = setInterval(() => {
      console.log(this.progressValue);
      this.progressValue++;
      if (this.progressValue > 100) {
        clearInterval(intervalId);
      }
    }, 250);
  }
}
```

I'm using `outputType: 'component'`, this is the angular output target config:

```ts
angular({
  componentCorePackage,
  outputType: 'component', // Generate many component wrappers tied to a single Angular module (lazy/hydrated approach)
  directivesProxyFile: resolve(__dirname, '../beeq-angular/src/directives/components.ts').replace(/\\/g, '/'),
  directivesArrayFile: resolve(__dirname, '../beeq-angular/src/directives/index.ts').replace(/\\/g, '/'),
  valueAccessorConfigs: angularValueAccessorBindings,
  customElementsDir,
}),
```

Here 👉🏼  is [a code sandbox with the code example](https://codesandbox.io/p/devbox/flamboyant-matan-3ykkm9?file=%2Fsrc%2Fapp%2Fapp.component.ts)

alexandertrefz:
@dgonzalezr This bug is specific to the new experimental `inlineProperties` configuration option, for the output, so in your case, update the output target config like this:

```typescript
angular({
  componentCorePackage,
  outputType: 'component', // Generate many component wrappers tied to a single Angular module (lazy/hydrated approach)
  directivesProxyFile: resolve(__dirname, '../beeq-angular/src/directives/components.ts').replace(/\\/g, '/'),
  directivesArrayFile: resolve(__dirname, '../beeq-angular/src/directives/index.ts').replace(/\\/g, '/'),
  valueAccessorConfigs: angularValueAccessorBindings,
  customElementsDir: 'dist/components',
  inlineProperties: true
})
```

dgonzalezr:
> @dgonzalezr This bug is specific to the new experimental `inlineProperties` configuration option, for the output, so in your case, update the output target config like this:

Thank you @alexandertrefz for clarifying it 🙌🏼 , I'll test the new experimental `inlineProperties`

kristilw:
Looking at it more I found that it is unrelated to `outputType: 'standalone'` and `type: 'dist-custom-elements'`, but probably the tsconfig of the Angular build. In the example angular project, the inlined props are removed from the final js build. @alexandertrefz, would you care to try the tsconfig as it is set up in the example project ([link](https://github.com/ionic-team/stencil-ds-output-targets/blob/main/example-project/component-library-angular/tsconfig.json))?

If using getters and setters resolves the problem that could be a solution, but it would be even better if they are never included in the final output (leaving the generated js the same regardless of the value of `inlineProperties`).

dgonzalezr:
> Playing around with the target setting in typescript, I found that by setting the `compilerOptions.target: 'es2022'` recreates the issue. However, the issue can then be resovled by setting another property in the config: `compilerOptions.useDefineForClassFields: false`. Care to test?
> 
> Not saying that this is a perfect solution, but just to verify

@kristilw I have applied the following changes based in your comment, and indeed worked for me with the new experimental config 🤷🏼‍♂️ 

CleanShot 2024-11-01 at 14 11 15

CleanShot 2024-11-01 at 14 12 13

I have no doubt that @alexandertrefz solutions could be the right approach, I'm just leaving some feedback so that all cases can be considered 🙂

alexandertrefz:
Now I had the time to test this similarly. I have indeed been using `ES2022` as my `target`. I could have sworn that I tried `useDefineForClassFields` before I figured out the proposed solution of inlining the Getters & Setters.

But indeed, `useDefineForClassFields: false` also seems to be working for me.

I am undecided which approach is better, on one hand, just changing the compilation options seems like a cleaner solution, that has less code duplication, etc.
On the other hand, it seems slightly more cryptic to depend on the compilation settings to make this work, and should be put into the docs, at the very least.

At the very least it is good to know that it is possible to make it work, with the current version! Thanks @kristilw!

christian-bromann:
@alexandertrefz mind rebasing the PR?

alexandertrefz:
@christian-bromann Yes, apologies, I got quite tied up with work. I'll update & rebase it later today.

alexandertrefz:
@christian-bromann Rebased it again, if you can run the test, it should be good to go to merge.

christian-bromann (approved):
LGTM 👍

alexandertrefz:
Should I rebase the branch one more time? The tests ran through clean.

christian-bromann:
@alexandertrefz no need, thanks! I will publish a release next week. Thanks a lot!

PR #574, opened 2024-12-17
fix(vue): import JSX from default component library export

christian-bromann:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

In Vue the output target would generate a file importing the JSX namespace from `/components`, e.g.

```ts
import type { JSX } from 'component-library/components';
```

This works when using the old Node classic import resolution but fails with ESM and Node16 and up.

## What is the new behavior?

Since the the default export just re-exports what's in `/components`, e.g. `export * from './components';` we can just adjust the import path.

- [ ] Yes
- [x] No

### DIFF
--- a/example-project/component-library-react/package.json
+++ b/example-project/component-library-react/package.json
@@ -25,7 +25,7 @@
     "@typescript-eslint/eslint-plugin": "^8.8.1",
     "@typescript-eslint/parser": "^8.8.1",
     "@vitejs/plugin-react": "^4.3.2",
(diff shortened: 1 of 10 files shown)


tanner-reits (approved):
Only thing I can think that might have an impact on this is needing the `single-export-module` export type on the CE build, but we have that called out as a requirement in the docs when using `includeImportCustomElements`

PR #641, opened 2025-05-19
fix(ssr): improved SSR handling in Next.js

christian-bromann:
## Pull request checklist

This PR polishes the SSR process within Next.js applications and elsewhere. It addresses the following issues:

#### Variable Support

Before passing it property values as variables would not work, e.g.:

```tsx
// ❌ variable would not be resolved
const val = "hello world"
return <MyComponent prop={val} />
```

We now properly resolve the value if defined in the right scope, however this still will fail if the value is dynamically generated as we don't have access to its value and compile time, e.g.:

```tsx
// ✅ variable is being resolved
const val = "hello world"
// ❌ variable resolved at runtime will still fail
const otherVal = someFunction()
return <MyComponent prop1={val} prop2={otherVal} />
```

#### Transform Components from other sources

If a component is imported from a different location (outside of the `pages` or `app` folder), it wouldn't be transformed properly, removing all SSR support for these components. This has been fixed by adjusting the Webpack loader.

#### Support for Spread Properties

Before using spread properties would cause a failure, e.g.:

```tsx
// ❌ this would fail
const props = { first: "Hello", last: "World" }
return <MyComponent {...props} />
```

With this patch, it is now supported.

#### Support for Non-Dev Environments

In previous versions we would only check for `jsxDev` calls on Stencil components to do the right AST transformation. This has been extended and also supports production `jsx` calls.

#### Reduction of Hydration Errors in Next.js

Using `@stencil/ssr` in Next.js would still cause Hydration errors given that the `template` tag of the DSD would be removed (and replaced with a Shadow DOM). This has caused hydration errors and potential re-rendering of the component, destroying the ShadowRoot completely.

This is now being resolved by using the `dynamic` primitive of Next.js to tell it to replace the SSR rendered component with the runtime version of it, e.g. we modify the source file to wrap the component as following:

```ts
import dynamic from 'next/dynamic';
const compImport = import('component-library-react');

const getMyInput$0 = ({ children }) => dynamic(
  () => compImport.then(mod => mod.MyInput),
  {
    ssr: false,
    loading: () => (<>
      <my-input class="hydrated" s-id="17">
        <template shadowrootmode="open" shadowrootdelegatesfocus="true" dangerouslySetInnerHTML={{ __html: `<input aria-labelledby="my-input-0-lbl" autocapitalize="off" autocomplete="off" autocorrect="off" c-id="17.0.0.0" class="native-input" name="my-input-0" placeholder="" type="text" value="">` }}></template>
        {children}
      </my-input>
    </>)
  }
)
```

⚠️ Warning: this still causes issues when your components use slots as Next.js injects `<template data-dgst="BAILOUT_TO_CLIENT_SIDE_RENDERING" data-msg="Bail out to client-side rendering: next/dynamic" data-stck="..." />` which (afaik) can't be disabled causing your component to deal with undesired child nodes, e.g.:

```tsx
<MyStencilComponent>
  <template ...>
  <ExpectedSlotNode />
  <template ...>
  <ExpectedSlotNode />
</MyStencilComponent>
```

While the template tag doesn't render anything, it may be picked as slot component, causing issues in rendering.

- [ ] Build (`npm run build`) was run locally for affected output targets
- [ ] Prettier (`npm run prettier`) was run locally and passed

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

- [ ] Yes
- [x] No

## Other information

Updated SSR docs: https://github.com/stenciljs/site/pull/1525


example-project/next-15-react-19-app/test/test.e2e.ts
```diff
@@ -3,10 +3,12 @@ import { runTestScenarios } from 'react-test-components/scenarios';
...
-      'complex-props-scoped',
-      'complex-props-shadow',
-      'single-children-shadow',
-      'nested-shadow'
+      /**
+       * currently broken in Next.js
+       *
+       * Problem here is that React doesn't re-render the component at runtime to initate the event handlers.
+       */
+      'input-scoped'
```
christian-bromann, line 11: This is good news 🎉

example-project/react-app/tests/test.e2e.ts
```diff
@@ -1,5 +1,5 @@
 import { runTestScenarios } from 'react-test-components/scenarios';
 
-describe('React Vite SSR Integration', () => {
+describe.skip('React Vite SSR Integration', () => {
```
christian-bromann, line 3: This not so much 😞

christian-bromann:
__Update:__

I spend more time trying to update for Next.js applications and made some progress. This has become much more complex as anticipated. Unfortunately while we see improvements in Next.js, we see now issues in Remix and React+Vite. I will take another stab at this tomorrow and next week. In the meantime I would appreciate feedback especially if you use Next.js. A dev build is available via:

```sh
npm i @stencil/ssr@0.0.1-dev.11747956973.17340fd0
```

Also I updated the docs and published a preview: https://stencil-docs-git-cb-ssr-doc-improvements-ionic1.vercel.app/docs/next/server-side-rendering

Issue #646, opened 2025-06-02
bug: provide esm import file


davidpett:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/stenciljs/output-targets/issues) that already report this problem, without success.

### Stencil Version

4.31.0

### Stencil Framework Output Target

React

### Stencil Framework Output Target Version

1.0.3

### Current Behavior

when using `esModules: true` in stencil.config.ts there is no single file that lists the individual exports like there is in components.ts when `esModules: false`.

When publishing an npm package, we want to allow the user to use named imports:
```
import { MyButton } from '@my-lib/ui/react'
```

but with the current setup this isn't possible

### Expected Behavior

expect to be able to use esm in our component library react exports

### Steps to Reproduce

in the example-project/component-library/stencil.config.ts in this repo, add `esModules: true` to the reactOutputTarget config and run the build

### Code Reproduction URL

h

christian-bromann:
Fixed in https://github.com/stenciljs/output-targets/pull/647

PR #647, opened 2025-06-03
fix(#646): react create components.ts file

davidpett:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

esModules: true in react-output-target does not create a single file that exports all components which makes it hard to create an npm package with an entry point. 

Issue URL:
https://github.com/stenciljs/output-targets/issues/646
## What is the new behavior?

- this ensures that there is a components.js file created that maps to each esm file

- [ ] Yes
- [x] No


packages/react/src/create-component-wrappers.ts
```diff
@@ -1,6 +1,7 @@
-import path from 'node:path';
 import type { ComponentCompilerMeta } from '@stencil/core/internal';
+import path from 'node:path';
```
christian-bromann, line 2: Let's keep Node.js imports first

package.json
```diff
@@ -13,6 +13,7 @@
     "build.example.angular": "pnpm --filter component-library-angular run build",
     "build.example.core": "pnpm --filter component-library run build",
     "build.example.react": "pnpm --filter component-library-react run build",
+    "build.example.react-esm": "pnpm --filter component-library-react-esm run build",
```
christian-bromann, line 16: I think having the unit tests is sufficient for this. I want to avoid having too many integration tests as they slow down the pipeline significantly.

christian-bromann (changes requested):

package.json
davidpett: updated

packages/react/src/create-component-wrappers.ts
davidpett: oops, was editor formatting on save. fixed

package.json
davidpett: Ah, you want to delete the entire package? Got it. Will do

christian-bromann (commented):
Can we extend the `packages/react/src/create-component-wrappers.test.ts` tests to verify that with `esModules` set to `true` it also generates the `components.ts` file?

christian-bromann (approved):
LGTM 👍

PR #777, opened 2026-03-31
fix(react): use component type in component definitions

OS-jacobbell:
## Pull request checklist

## Pull request type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

1. A Stencil element extends a component and `HTMLElement`. E.g. `interface MyButton extends Components.MyButton, HTMLElement {}`.
2. A `StencilReactComponent` is templated on a Stencil element. E.g. `MyButton: StencilReactComponent<MyButtonElement, MyButtonEvents>`
3. `StencilReactComponent` is a `React.FunctionComponent` templated on `StencilProps`.
4. `StencilProps` contains `React.HTMLAttributes` and `Partial<Omit<I, keyof HTMLElement>>` where `I` is the Stencil element. E.g. `MyButtonElement`, which extended `HTMLElement`, now has the props of `HTMLElement` removed.

**Problem:** If the Stencil component uses a prop that is in `HTMLElement` and is not in `React.HTMLAttributes`, the prop is removed from the `StencilReactComponent`. This is blocking Ionic Framework from updating its Stencil react-output-target as components such as ion-input define `autofocus`, which is being removed by this typing.

Issue URL: https://github.com/stenciljs/output-targets/issues/475

## What is the new behavior?
Rather than omitting `keyof HTMLElement` from the Stencil element to find what was in the component, pass through the component directly. E.g. `MyButton: StencilReactComponent<MyButtonElement, Components.MyButton, MyButtonEvents>`. `Partial<Omit<I, keyof HTMLElement>>` is replaced by `Partial<C>`, where `C` is `Components.MyButton`.

- [ ] Yes
- [X] No

### DIFF
--- a/example-project/component-library-react/src/components.server.ts
+++ b/example-project/component-library-react/src/components.server.ts
@@ -162,11 +162,11 @@ export const MyComponent: StencilReactComponent<MyComponentElement, MyComponentE
 
 export type MyComponentDelegatesFocusEvents = NonNullable<unknown>;
 
-export const MyComponentDelegatesFocus: StencilReactComponent<MyComponentDelegatesFocusElement, MyComponentDelegatesFocusEvents> = /*@__PURE__*/ createComponent<MyComponentDelegatesFocusElement, MyComponentDelegatesFocusEvents>({
+export const MyComponentDelegatesFocus: StencilReactComponent<MyComponentDelegatesFocusElement, MyComponentDelegatesFocusEvents, Components.MyComponentDelegatesFocus> = /*@__PURE__*/ createComponent<MyComponentDelegatesFocusElement, MyComponentDelegatesFocusEvents, Components.MyComponentDelegatesFocus>({
     tagName: 'my-component-delegates-focus',
     properties: {},
     hydrateModule: import('component-library/hydrate') as Promise<HydrateModule>,
-    clientModule: clientComponents.MyComponentDelegatesFocus as ReactWebComponent<MyComponentDelegatesFocusElement, MyComponentDelegatesFocusEvents>,
+    clientModule: clientComponents.MyComponentDelegatesFocus as StencilReactComponent<MyComponentDelegatesFocusElement, MyComponentDelegatesFocusEvents, Components.MyComponentDelegatesFocus>,
     serializeShadowRoot,
     getTagTransformer
 });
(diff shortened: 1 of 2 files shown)


packages/react/src/runtime/create-component.ts
```diff
@@ -1,35 +1,45 @@
 import type { EventName, Options } from '@lit/react';
 import { createComponent as createComponentWrapper } from '@lit/react';
+import { ComponentInterface } from '@stencil/core';
```
ShaneK: This imports `ComponentInterface` as a value but it's a pure interface, right? Shouldn't this be `import type { ComponentInterface } from '@stencil/core'`? [Same thing in `ssr.tsx`](https://github.com/stenciljs/output-targets/pull/777/files/6a8e32c998a96e7dc3dfcbdc95590f9e3f3784cf#diff-202611a57fef176f0bd09f8ee324b4e5516a0fd072cca69c8fcb6d56ea77810aR7). The other imports in this file use `import type` for type-only stuff, and the generated code in the code generator [already does it correctly](https://github.com/stenciljs/output-targets/pull/777/files/6a8e32c998a96e7dc3dfcbdc95590f9e3f3784cf#diff-fd20b36304fcceec73fff5943d025356520cf2da69109acad43e101787acfb3fR61) with `import type { Components }`. Without it you'd get a runtime dependency on `@stencil/core` in the shipped bundle.

packages/react/src/runtime/ssr.tsx
```diff
@@ -449,7 +451,7 @@ export const createComponent = <I extends HTMLElement, E extends EventNames = {}
       return createComponentWrapper<I, E>({
```
ShaneK: This still says `createComponentWrapper<I, E>` but after [the changes to `createComponent`](https://github.com/stenciljs/output-targets/pull/777/files/6a8e32c998a96e7dc3dfcbdc95590f9e3f3784cf#diff-7bf9d7de5586b4c771b453dca07565ae83df5bd7185edaededf5a6690fe1edcaR27) it now takes 3 type params `<I, C, E>`. TypeScript fills these left-to-right so `E` ends up in the `C` slot and the actual events default to `{}`. The casts mask it at compile time but shouldn't this be `createComponentWrapper<I, C, E>` to match the updated signature?

PR #785, opened 2026-04-15
chore(tests): test for nextjs hydration mismatch errors

OS-jacobbell:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [X] Other (please describe): test cases

All tests in example projects ignore hydration mismatch errors.

## What is the new behavior?

Integrated the lighter-weight SSR tests by OS-paulvisciano.
- Added components in NextJS tests.
- Enable client side error detection in NextJS tests.
- Explicitly ignore hydration mismatch errors in react-app and remix-app tests. They still have hydration mismatch errors and hydration mismatches are no longer ignored across the board.

- [ ] Yes
- [X] No


tests/react/tests/scenarios.ts
```diff
@@ -38,11 +37,18 @@ export const runTestScenarios = ({ only, exclude, ...opts }: TestScenarioOptions
...
+      // Set the current test component so assertClientSideErrors can identify known failures
+      // We set it in beforeEach and keep it set so afterEach hooks can access it
+      beforeEach(() => {
+        ;(global as any).__currentTestComponent = name
+      })
+      test()
+      // Note: We don't clear __currentTestComponent here because the afterEach
+      // hook in helpers.ts needs to access it, and hooks run in reverse order
+    })
   })
```
ShaneK:
```suggestion
  scenarions.forEach(([name, test]) => {
    describe(name, test)
  })
```

`assertClientSideErrors` only reads the `ignoreHydrationMismatchErrors` boolean from its closure. It never reads `__currentTestComponent`. The `beforeEach` write here and the comments around it ("so `assertClientSideErrors` can identify known failures", "the afterEach hook in helpers.ts needs to access it") describe a per-component known-failures wiring that isn't actually in place. Was the per-component idea dropped in favor of the suite-level boolean and this got left behind, or is the read side meant to land separately?

tests/react/tests/helpers.ts
```diff
@@ -117,23 +121,44 @@ export function assertClientSideErrors() {
    */
   afterEach(() => {
     /**
-     * TODO: include all errors once we have fixed the hydration issues
+     * Filter errors based on whether this is a known failure
+     * For known failures, we ignore hydration errors
+     * For other components, we check for hydration errors
```
ShaneK: The JSDoc here and the inline comments at lines 138 (`// If it's a known failure, ignore hydration errors for this component`) and 156 (`// For hydration errors on non-known-failures, we want to catch them`) all describe a per-component decision, but the function takes one boolean that's applied uniformly to every error from every component in the suite. There's no "this component" / "known failure" branching anywhere in the body. Could you rewrite these to match the actual suite-wide behavior? Otherwise a future reader is going to look for the per-component dispatch and not find it.

tests/react/src/next/TestComponent.tsx
```diff
@@ -40,6 +46,18 @@ const testComponents = [
...
+  // individual component tests
+  'my-button',
+  'my-checkbox',
+  'my-component',
+  'my-counter',
+  'my-list',
+  'my-popover',
+  'my-radio-group',
+  'my-range',
+  'my-toggle',
```
ShaneK: The 10 entries under `// individual component tests` get added to the `testComponents` type union and pick up render branches below, but I don't see matching scenario keys in `tests/react/tests/scenarios/{scoped,shadow,transformed}.ts`. `runTestScenarios` only iterates keys from those records, so the new pages at `next-15-runtime-based/src/app/my-*/page.tsx` never get visited by the e2e harness. Are the actual `it` blocks coming in a follow-up PR, or did the wiring get missed here?

example-project/next-app/test/test.e2e.ts
```diff
@@ -2,6 +2,7 @@ import { runTestScenarios } from 'react-test-components/scenarios';
 
 describe('Next.js 14 React 18 SSR Integration', () => {
   runTestScenarios({
+    ignoreHydrationMismatchErrors: true,
```
ShaneK: This silences every hydration error in the next-app suite without naming which component(s) are actually failing or linking a tracker. The `next-15-runtime-based` file documents its excludes inline (`Known hydration mismatch errors` on the two scoped excludes), could you do the same here, ideally with an issue number to remove the flag against? Without that, when the underlying mismatches are fixed there's nothing pulling the flag back out.

tests/react/tests/scenarios.ts
```diff
@@ -38,11 +37,18 @@ export const runTestScenarios = ({ only, exclude, ...opts }: TestScenarioOptions
    * track all errors that are logged during the tests and assert
    * that no errors were logged
    */
-  if (typeof opts.assertClientSideErrors !== 'boolean' || opts.assertClientSideErrors) {
-    assertClientSideErrors()
-  }
+  assertClientSideErrors(!!opts.ignoreHydrationMismatchErrors);
```
ShaneK:
```suggestion
  assertClientSideErrors(!!opts.ignoreHydrationMismatchErrors)
```

Tiny nit, this is the only line in the file with a trailing semicolon, so suggesting this just for consistency. I'm not a huge fan of semi-colon-less JS, but I am a fan of consistency I guess haha

ShaneK (changes requested):
I had some questions and noticed a few issues (most pretty minor!), but great work so far!

tests/react/tests/scenarios.ts
OS-jacobbell: There was both suite-level and per-component. Per-component was only used by the next-15-runtime-based project and overlapped confusingly with the exclusion list, so I reduced to just suite-level but forgot to simplify this part.

tests/react/src/next/TestComponent.tsx
OS-jacobbell: Actually added these components to the spec files and "example-project" test apps.

example-project/next-app/test/test.e2e.ts
OS-jacobbell: Previously hydration mismatch errors were ignored globally, so this boolean is applied to all the projects besides the focus of these tests, which was next-15-runtime-based. However, most of the components are hydration-mismatch-free in next-app, so I removed the ignore boolean on this one.

tests/react/tests/helpers.ts
```diff
@@ -116,24 +120,39 @@ export function assertClientSideErrors() {
...
+        // Filter out WebdriverIO related socket errors
+        if (
+          error.includes('failed to connect to websocket') ||
+          error.includes('WebSocket closed without opened.')
+        ) {
+          return false;
+        }
+        // Keep other non-hydration errors
+        return true;
+      }
```
ShaneK, line 146: This comment still talks about per-component "known failures," but the function only takes a suite-wide `ignoreHydrationMismatchErrors` boolean and there's no per-component branching anywhere in the body. Same family of issue you replied to last round (the `__currentTestComponent` global is gone, but this comment got missed). Could you reword to something like `// Hydration errors here weren't opted out of, so they're real failures`?

example-project/react-app/tests/test.e2e.ts
```diff
@@ -2,6 +2,7 @@ import { runTestScenarios } from 'react-test-components/scenarios';
 
 describe('React Vite SSR Integration', () => {
   runTestScenarios({
+    ignoreHydrationMismatchErrors: true,
```
ShaneK, line 5: Same feedback you addressed for `next-app`: this swallows hydration errors across the whole suite, including for tests that are currently passing, so a regression in any of them won't fail CI. Could you convert this to per-component `exclude` entries the way `next-app` and `next-15-runtime-based` ended up, and name which components are actually broken (ideally with a tracker link to remove the flag against)?

example-project/remix-app/tests/test.e2e.ts
```diff
@@ -2,6 +2,7 @@ import { runTestScenarios } from 'react-test-components/scenarios';
 
 describe('Remix SSR Integration', () => {
   runTestScenarios({
+    ignoreHydrationMismatchErrors: true,
```
ShaneK, line 5: Same as in [`react-app/tests/test.e2e.ts`](https://github.com/stenciljs/output-targets/pull/785/files/f3306d1b70e27a0916da9ac22ec5477d63847a60#diff-f45d3a509071b0fa991476657ffff84961f103aa42143fa80392bd2abe126074R5). Per-component excludes with named failing components is going to age better than a suite-wide flag here. Otherwise hydration regressions in the currently-passing scenarios go silent.

tests/react/src/TestComponent.tsx
```diff
@@ -165,6 +177,40 @@ const TestComponent = ({ name }: TestComponentProps) => {
       </>
     );
   }
+  if (name === 'my-button-scoped') {
+      return <MyButtonScoped>Click Me</MyButtonScoped>;
+    }
```
ShaneK:
```suggestion
  if (name === 'my-button-scoped') {
    return <MyButtonScoped>Click Me</MyButtonScoped>;
  }
```

The indentation in this whole block is off compared to the rest of the function. The first `if` is at 2-space but its body is at 6-space, and every following `if` is at 4-space. The parallel block in `tests/react/src/next/TestComponent.tsx` has the same logic correctly at 2-space throughout. `pnpm prettier --write` on this file should clean up the rest.

example-project/next-15-runtime-based/src/app/my-button-shadow/page.tsx
```diff
@@ -0,0 +1,9 @@
+import { TestStage } from 'react-test-components/next';
+
+export default function MyButtonScopedPage() {
```
ShaneK:
```suggestion
export default function MyButtonShadowPage() {
```

Copy-paste leftover, should be `MyButtonShadowPage` to match the route (the body renders `my-button-shadow`). Next ignores the function name so it doesn't break anything, but the parallel `remix-app/app/routes/my-button-shadow.tsx` already uses `MyButtonShadow`.

tests/react/tests/scenarios/shadow.ts
```diff
@@ -86,5 +86,47 @@ export const testScenarios: Record<ShadowComponents, () => void> = {
         'Shadow Change Event: Hello World!',
       ].join('\n'))
     })
-  }
+  },
+  'my-button-shadow': () => {
+    it('should not have hydration mismatch errors', async () => {
```
ShaneK: The `it` description promises a hydration check, but the body is just a presence check. The actual hydration assertion comes from the global error listener in `helpers.ts`, which is non-obvious. Could you rename to `'should render without errors'` or add a comment pointing at where the hydration check actually lives? Same applies to all the new shadow scenarios in this hunk and the new `'my-button-scoped'` scenario in `scenarios/scoped.ts`.

ShaneK (changes requested):
Solid improvements, but still some issues

example-project/react-app/tests/test.e2e.ts
OS-jacobbell: Half the test components log "hydration mismatch error" in this and the remix app, but work fine otherwise, including server side rendering. Previously, hydration mismatch checking was disabled globally. This PR moves it to a per-app flag and focuses on test coverage just for NextJS.

tests/react/tests/helpers.ts
OS-jacobbell: Right that makes sense, I think it really doesn't needs comments there at all.

tests/react/src/next/TestComponent.tsx
```diff
@@ -40,6 +43,16 @@ const testComponents = [
   'transform-scoped-to-shadow',
   'transform-tag-test',
   'style-deduplication-scoped',
+  'style-no-deduplication-scoped',
```
ShaneK: The regular `tests/react/src/TestComponent.tsx` excludes this scenario from `ScopedComponents` (line 61) and adds it to `TransformedComponents` (line 65), but the type aliases below on lines 59-66 of this file weren't updated to match. So `style-no-deduplication-scoped` falls into `ScopedComponents` here, even though `scenarios/transformed.ts:129` is what actually implements it. Want to sync the two `TestComponent.tsx` files?

example-project/component-library/src/components/my-button/my-button.tsx
```diff
@@ -17,7 +17,7 @@ import { AnchorInterface, ButtonInterface } from '../element-interface';
   styleUrl: 'my-button.css',
   shadow: true,
 })
-export class Button implements ComponentInterface, AnchorInterface, ButtonInterface {
+export class MyButton implements ComponentInterface, AnchorInterface, ButtonInterface {
```
ShaneK, line 20: The "make component names consistent" pass also touched `my-dialog/` (renamed to `my-popover/`), but [`my-popover.tsx:14`](https://github.com/stenciljs/output-targets/blob/5d9c277ea1644b89d4f6c4eba15f65621f54755f/example-project/component-library/src/components/my-popover/my-popover.tsx#L14) still reads `export class Popover` while its element type is already `HTMLMyPopoverElement`. Want to rename it to `MyPopover` to match the rest of the renames in this commit?

ShaneK (approved):
Looking good! Just left some nits. Fix them if you want

PR #788, opened 2026-04-22
fix(react): reflect required props in component type when used in TSX

axelfriberg:
## Pull request checklist

## Pull request type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

When setting a prop to be required with exclamation mark (!) and then generating React components that is not reflected in the props when using the component in TSX. So `@Prop() label!: string;` `<MyButton />` doesn't show an error. 

## What is the new behavior?

This PR fixes so that it gives a type error if required props are not provided. It does so with the same approach that the Vue component generator does by using `JSX.MyButton` exported by components.d.ts.

- [ ] Yes
- [x] No

## Other information

There is no issue created but discussed in Discord https://discord.com/channels/1347590794130755665/1493565613896110161

### DIFF
--- a/packages/react/src/runtime/create-component.ts
+++ b/packages/react/src/runtime/create-component.ts
@@ -5,40 +5,39 @@ import { createComponent as createComponentWrapper } from '@lit/react';
 type EventNames = Record<string, EventName | string>;
 
 // Type that's compatible with both React 18 and 19
-type StencilProps<I extends HTMLElement, E extends EventNames, C> = Omit<
-  React.HTMLAttributes<I>,
-  keyof E
+type StencilProps<Element extends HTMLElement, Events extends EventNames, Props> = Omit<
+  React.HTMLAttributes<Element>,
+  keyof Events
 > &
-  Partial<{ [K in keyof E]: E[K] extends EventName<infer T> ? (event: T) => void : (event: any) => void }> &
-  Partial<C> &
-  React.RefAttributes<I>;
+  Partial<{ [K in keyof Events]: Events[K] extends EventName<infer T> ? (event: T) => void : (event: any) => void }> &
+  Props &
+  React.RefAttributes<Element>;
 
 export type StencilReactComponent<
-  I extends HTMLElement,
-  E extends EventNames = {},
-  C = Omit<I, keyof HTMLElement>,
-> = React.FunctionComponent<StencilProps<I, E, C>>;
+  Element extends HTMLElement,
+  Events extends EventNames = {},
+  Props = {},
+> = React.FunctionComponent<StencilProps<Element, Events, Props>>;
 
 /**
  * Defines a custom element and creates a React component.
  * @public
  */
-export const createComponent = <I extends HTMLElement, E extends EventNames = {}, C = Omit<I, keyof HTMLElement>>({
+export const createComponent = <Element extends HTMLElement, Events extends EventNames = {}, Props = {}>({
   defineCustomElement,
   tagName,
   transformTag,
   ...options
-}: Options<I, E> & {
+}: Options<Element, Events> & {
   defineCustomElement: () => void;
   transformTag?: (tagName: string) => string;
-}): StencilReactComponent<I, E, C> => {
+}): StencilReactComponent<Element, Events, Props> => {
   if (typeof defineCustomElement !== 'undefined') {
     defineCustomElement();
   }
   const finalTagName = transformTag ? transformTag(tagName) : tagName;
-  return createComponentWrapper<I, E>({ ...options, tagName: finalTagName }) as unknown as StencilReactComponent<
-    I,
-    E,
-    C
-  >;
+  return createComponentWrapper<Element, Events>({
+    ...options,
+    tagName: finalTagName,
+  }) as unknown as StencilReactComponent<Element, Events, Props>;
 };
--- a/example-project/component-library-angular/projects/library/src/directives/proxies.ts
+++ b/example-project/component-library-angular/projects/library/src/directives/proxies.ts
@@ -204,7 +204,7 @@ export declare interface MyComplexPropsScoped extends Components.MyComplexPropsS
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: '<ng-content></ng-content>',
   // eslint-disable-next-line @angular-eslint/no-inputs-metadata-property
-  inputs: ['first', 'last', 'middleName'],
+  inputs: [{ name: 'first', required: true }, 'last', 'middleName'],
 })
 export class MyComponent {
   protected el: HTMLMyComponentElement;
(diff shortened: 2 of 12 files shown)


packages/react/src/runtime/create-component.ts
```diff
@@ -5,40 +5,39 @@ import { createComponent as createComponentWrapper } from '@lit/react';
...
+  React.RefAttributes<Element>;
 
 export type StencilReactComponent<
-  I extends HTMLElement,
-  E extends EventNames = {},
-  C = Omit<I, keyof HTMLElement>,
-> = React.FunctionComponent<StencilProps<I, E, C>>;
+  Element extends HTMLElement,
+  Events extends EventNames = {},
+  Props = {},
```
johnjenkins, line 19: removing `C = Omit<I, keyof HTMLElement>,`
adding `Props = {},`
can you say something about this change? 🙏

packages/react/src/runtime/create-component.ts
```diff
@@ -5,40 +5,39 @@ import { createComponent as createComponentWrapper } from '@lit/react';
...
+> = React.FunctionComponent<StencilProps<Element, Events, Props>>;
 
 /**
  * Defines a custom element and creates a React component.
  * @public
  */
-export const createComponent = <I extends HTMLElement, E extends EventNames = {}, C = Omit<I, keyof HTMLElement>>({
+export const createComponent = <Element extends HTMLElement, Events extends EventNames = {}, Props = {}>({
```
johnjenkins, line 26: Same question :)

packages/react/src/runtime/create-component.ts
axelfriberg: Sure! When renaming the generics to make it a bit more clear what they are rather than <I, E, C> and align with the Vue generics I thought that `= {}`  might be a bit of clearer default since that would equal the same result as what happens when you generate a component with no props, which is an empty interface in the JSX namespace. But I was also thinking about maybe we might as well just remove the defaults? Since they are always set by the generator script. 

So very much my opinion stuff so if you prefer/think it is more correct to have `Props = Omit<Element, keyof HTMLElement>` I could change it back to be that instead :)

johnjenkins (approved):

ShaneK:
@johnjenkins this PR caused a massive regression in 1.5.1 by removing the `Partial<C>` wrapper, [making you define everything](https://github.com/ionic-team/ionic-framework/actions/runs/25007199736/job/73233599812?pr=30831)

The PR description actually points at the right fix: Vue references `JSX.MyButton` from `components.d.ts`, not `Components.MyButton`. That's the shape that lines up with runtime behavior, because Stencil emits `?` on optionals in the `JSX` namespace but leaves them required in the `Components` namespace. React's third type parameter is still wired to `Components.X`, so every default-valued prop is now required in TSX

Could you switch the third type parameter from `Components.X` to `JSX.X` to match Vue, or restore the `Partial<C>` wrapper? Either path keeps the original goal of erroring on genuinely required props

ShaneK:
Hey @johnjenkins, no worries. The `Components.*` references are in the generated `components.ts` file that ships inside `@ionic/react`, not in your runtime source.

The generator change from `Components.X` to `JSX.X` landed in this same PR, and that piece is correct. The break is in the seam between versions: a `components.ts` file generated by the `1.5.0` generator (which still emits `Components.X`) paired with the `1.5.1` runtime (which dropped `Partial<C>`). The `1.5.0` runtime tolerated `Components.X` because `Partial<C>` flipped its required-by-default fields into optional ones at the JSX call site. [With `Partial<C>` gone](https://github.com/stenciljs/output-targets/pull/788/files#diff-7bf9d7de5586b4c771b453dca07565ae83df5bd7185edaededf5a6690fe1edcaL13), every `Components.X` prop is now required.

We're pinning to `1.5.0` on our side until this is sorted, but every Stencil consumer with a library already on npm is going to hit the same wall the moment a downstream user does a fresh install, and we need this fixed so we can unpin

ShaneK:
Thank you so much @johnjenkins !

axelfriberg:
No worries with reverting this since it caused issues @johnjenkins ! I would like to see if we can figure out a way to make this work since we require it (pun intended) in my team and it would also align with how the Vue and Angular output targets are at the moment which would be good I think. 

I've been reading your input @ShaneK and have a bit of hard time understanding what goes wrong, mostly since I'm not that familiar with the Stencil/Ionic internals yet.

> Could you switch the third type parameter from Components.X to JSX.X to match Vue

Isn't that what is being done here https://github.com/stenciljs/output-targets/pull/788/changes#diff-07ee854adc8030f5e42e1e5b8681e86cff909c0bd2e74071442a0c7266c0f7bdR45 in the generated output? Or is it somewhere else you mean it would need to be switched? 

> or restore the Partial<C> wrapper? Either path keeps the original goal of erroring on genuinely required props

During my testing, in order for the TSX to actually show required props as required I needed to remove Partial<C>  otherwise it didn't show. 

> The generator change from Components.X to JSX.X landed in this same PR, and that piece is correct. The break is in the seam between versions: a components.ts file generated by the 1.5.0 generator (which still emits Components.X) paired with the 1.5.1 runtime (which dropped Partial<C>). 

When would this happen that a runtime and generator doesn't have the same version? Are they not in the same package and would therefore follow the same version?

axelfriberg:
Ah I see then I get it. But then it has to be done in a major right? So that the consumer gets force up to "^2.0.0"? Or do you see a way around it with making it buildtime again like you wrote above @johnjenkins ?

axelfriberg:
Yeah I agree that this might not be valid enough to just have a major release for just this change. And I also agree that the best would be to make it cohesive someway again so we avoid this problem in the future. Since React 19 should have good support for web components now it might be possible to make a 2.0.0 version of react output target that drops support for react 18 and just supports react 19 and therefore might be able to also drop the @lit/react dep to make it build time again? 

I understand that there isn't scope to look at this right now. Like you say for now we can lock it to version 1.5.1. 

Do you think the output targets will need a major bump anyway when stencil v5 is released @johnjenkins ? Or can they stick on what ever major they are?

ShaneK:
Sorry, we're on quite different timezones, but I'm also afraid this may introduce breaking changes into consumers of Ionic Framework were it still published. We could have fixed this - I believe, in theory - by publishing a new build of Ionic Framework pinned to >1.5.1 and the react package and react router packages with the same pinning and - in theory - that would work. The problem then is any component library with a `@Prop() X!: T` declaration would suddenly require that prop in downstream apps' TSX. Apps that compile clean today against `@ionic/react` could surface new TS errors. That's a type-only break (runtime is unchanged), but for downstream developers it's still a "my app stopped compiling on a minor bump" experience we'd want to avoid in a non-major version.

If this is something we want to do and could be possible to coordinate, v9 of Ionic Framework is coming up and that would be the ideal time to do it, and it would need to be in a major for output targets definitely. Other than that, doing it right now is too breaking of a change.

ShaneK:
It absolutely is expected that breaking changes wouldn't be pushed out in non-major versions, yes. That was my stipulation for making the change. If you want to release output-targets 2.0.0 with the change, then yeah, that makes sense (but is also something that requires more planning than "I want to make this change now"). You can't arbitrarily break consumers just because in minor or patch versions.

I understand Stencil hasn't typically followed this pattern well, but we need reliability.

ShaneK:
Oh yeah, if we could do the fix without the other issue, then I think that would probably be fine to go into a patch version. I thought you were referring to my hypothetical scenario there where we hadn't reverted and things broke. In that scenario, I was more worried about disconnections between versions breaking user apps (it's easy to have a different version of @ionic/core, @ionic/react, @ionic/react-router by accident, for example). 

I'm not sure how required types are handled right now in Ionic for React, they could just be being ignored which I agree is a bad state of things. I'd probably have to talk to the team on figuring out how we'd want to proceed with that possibility, but I could see us justifying it being in a patch version too.

PR #790, opened 2026-04-23
chore(tests): add @method and dynamic @prop react tests

OS-jacobbell:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [X] Other (please describe): tests

Issue URL:

## What is the new behavior?

- Test that React component wrapper makes available `@Method` component methods.
- Test that React component wrapper re-renders component when a `@Prop` value changes.

- [ ] Yes
- [X] No


example-project/remix-app/tests/test.e2e.ts
```diff
@@ -2,6 +2,13 @@ import { runTestScenarios } from 'react-test-components/scenarios';
...
 describe('Remix SSR Integration', () => {
   runTestScenarios({
-    exclude: ['complex-props-scoped', 'style-deduplication-scoped'],
+    exclude: [
+      'complex-props-scoped',
+      'style-deduplication-scoped',
+      /**
+       * Known issue: updating a prop does not trigger a rerender.
+       */
+      'single-no-child-shadow'
```
ShaneK: Excluding the entire `single-no-child-shadow` scenario silences three tests that were green for remix: the SSR snapshot, the hydration check, and the attached-styles assertion. [`runTestScenarios` filters at scenario granularity](https://github.com/stenciljs/output-targets/blob/cb6f78694de54c4baa81cc9eba700dd85e33a0d5/tests/react/tests/scenarios.ts#L35), so there's no way to drop just the new prop-change `it()` from this list.

Could the prop-change assertion live in its own scenario like `prop-update-shadow` instead, so remix only excludes that one?

example-project/next-15-runtime-based/test/__snapshots__/test.e2e.ts.snap
```diff
@@ -1,5 +1,65 @@
 // Snapshot v1
 
+exports[`checkbox-shadow > should correctly server side render 1`] = `
+"
+<div id="checkbox-shadow">
+  <my-checkbox
+    aria-checked="undefined"
+    class="checkbox-label-placement-undefined hydrated interactive ios sc-my-checkbox sc-my-checkbox-h"
```
ShaneK: `aria-checked="undefined"` and `checkbox-label-placement-undefined` mean SSR isn't applying the defaults from [`my-checkbox.tsx`](https://github.com/stenciljs/output-targets/blob/cb6f78694de54c4baa81cc9eba700dd85e33a0d5/example-project/component-library/src/components/my-checkbox/my-checkbox.tsx#L82) (`labelPlacement = 'start'`, `checked = false`). The other three apps render `aria-checked="false"` and `checkbox-label-placement-start` correctly.

Snapshotting this locks the bug in as expected behavior. Should this be tracked as a known regression and the test skipped, or the underlying init fixed before baselining the markup?

example-project/next-app/test/__snapshots__/test.e2e.ts.snap
```diff
@@ -1,5 +1,72 @@
 // Snapshot v1
 
+exports[`checkbox-shadow > should correctly server side render 1`] = `
+"
+<div id="checkbox-shadow">
+  <template
+    data-dgst="BAILOUT_TO_CLIENT_SIDE_RENDERING"
```
ShaneK, line 7: The `BAILOUT_TO_CLIENT_SIDE_RENDERING` marker plus the `<div style="display:contents">` wrapper below mean Next.js fell back to client rendering for this subtree. The test name says "should correctly server side render" but the markup proves the SSR boundary bailed.

The next-15, react-app, and remix-app variants all SSR this case without that wrapper. Worth digging into what's throwing in next-14 (the `data-msg="..."` placeholder is masking the real reason), or renaming the test to match what's actually being asserted?

ShaneK (changes requested):
Looking good! Noticed a few issues with the tests, specifically the snapshots

example-project/remix-app/tests/test.e2e.ts
OS-jacobbell: Added a prop-update-shadow scenario.

example-project/next-15-runtime-based/test/__snapshots__/test.e2e.ts.snap
OS-jacobbell: Added it as a known regression and made it skip the test.

example-project/next-app/test/__snapshots__/test.e2e.ts.snap
OS-jacobbell: A couple other components have `BAILOUT` in the next-app with the same test name, so I don't know, it seems fine and does actually server side render.

tests/react/src/TestComponent.tsx
```diff
@@ -120,12 +123,18 @@ const TestComponent = ({ name }: TestComponentProps) => {
       />
     );
   }
+  if (name === 'prop-update-shadow') {
+      return <MyComponent first="John" middleName="William" last="Doe" />;
+    }
```
ShaneK:
```suggestion
  if (name === 'prop-update-shadow') {
    return <MyComponent first="John" middleName="William" last="Doe" />;
  }
```

Indentation is off by one level here, so the body and closing brace sit deeper than every other `if` branch in this file. The next variant at [`next/TestComponent.tsx:125-127`](https://github.com/stenciljs/output-targets/blob/dd1b3823c1d795d274ea2bfbe2b75fb3e78ef1aa/tests/react/src/next/TestComponent.tsx#L125-L127) already has the right shape.

tests/react/tests/scenarios/shadow.ts
```diff
@@ -77,6 +77,13 @@ export const testScenarios: Record<ShadowComponents, () => void> = {
       ].join('\n'))
     })
   },
+  'prop-update-shadow': () => {
+    it('should update rendered text when prop changes', async () => {
+      await browser.url('/single-no-child-shadow')
```
ShaneK, line 82: The scenario is `prop-update-shadow` and you've added a matching `if (name === 'prop-update-shadow')` branch in both [`TestComponent.tsx`](https://github.com/stenciljs/output-targets/blob/dd1b3823c1d795d274ea2bfbe2b75fb3e78ef1aa/tests/react/src/TestComponent.tsx#L126-L128) and [`next/TestComponent.tsx`](https://github.com/stenciljs/output-targets/blob/dd1b3823c1d795d274ea2bfbe2b75fb3e78ef1aa/tests/react/src/next/TestComponent.tsx#L125-L127), plus a `'prop-update-shadow'` entry in `testComponents`, but none of the apps add a `/prop-update-shadow` page route. So this navigates to `/single-no-child-shadow` and the new test-component branch is unreachable. Want to drop the dead mapping (the existing route already renders the same JSX), or add the route pages and switch the URL here?

tests/react/tests/scenarios/shadow.ts
```diff
@@ -77,6 +77,13 @@ export const testScenarios: Record<ShadowComponents, () => void> = {
...
+      await browser.execute((el: any) => { el.middleName = 'Test'; }, await $('my-component'))
+      await expect($('my-component').$('div')).toHaveText("Hello, World! I'm John Test Doe")
```
ShaneK, line 84: `browser.execute((el) => { el.middleName = 'Test' })` writes the property directly on the underlying Stencil element, so this never goes through React's render path. The PR description says this verifies "the React component wrapper re-renders component when a `@Prop` value changes", but the assertion would pass identically without a React wrapper in front. Was the intent to mutate React-side state (e.g., an interactive route where a button calls `setState`), or is exercising the underlying Stencil reactivity what you wanted here?

ShaneK (changes requested):

tests/react/tests/scenarios/shadow.ts
OS-jacobbell: Uhh, yeah that's fine. I'm not too sure where the line is between the Lit component and underlying Stencil component. I wanted to just check that a prop can be changed and trigger an update. This test does fail in the remix-app so it's doing its job.

tests/react/tests/scenarios/shadow.ts
OS-jacobbell: Merged with the main branch, which uses dynamic routes in the SSR apps just like the react-app. Now routes no longer have to be added manually for each test component.

ShaneK (approved):
Awesome work!

Issue #791, opened 2026-04-24
bug: React output target runtime resolves to node/SSR path under Vitest and drops custom element prop forwarding


mhoritani:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/stenciljs/output-targets/issues) that already report this problem, without success.

### Framework Output Target

React (@stencil/react-output-target)

### Output Target Version

1.2.0 (also reproduced with 1.5.1)

### Stencil Version

4.43.3

### Reproduction URL

~~TBD~~
https://github.com/mhoritani/stencil-react-vitest-prop-forwarding-repro

### Describe the Bug

In Vitest + jsdom, React wrappers generated by `@stencil/react-output-target` render the custom element tag, but element props are not forwarded in tests (for example `id` stays unset).

In our reproduction, `<gravity-button>` is rendered, and `customElements.get('gravity-button')` is defined, but `document.getElementById('hello-button')` returns `null` because `id` never gets attached.

This looks like runtime resolution selecting the node/SSR path for `@lit/react` in Vitest’s server-side transform context, so the browser prop-attachment behavior is skipped.

### Steps to Reproduce

1. Use a React app with Vitest + jsdom and `@testing-library/react`.
2. Install a Stencil-generated React wrapper package built with `@stencil/react-output-target` v1.x.
3. Render a wrapped component with a native prop, e.g. `<GravityButton id="hello-button" />`.
4. Assert `document.getElementById('hello-button')`.

Result: assertion fails because `id` is not attached.

### Expected Behavior

When rendering wrapped components in Vitest + jsdom, native element props (like `id`) should be forwarded and visible on the rendered custom element, same as in browser runtime.

### Additional Context

- Related issue: #615 (feature request about Vitest support).
- This issue is narrower: components render and are defined, but prop forwarding is dropped in Vitest.

Workaround that makes our repro pass:

```js
// vitest.config.js
export default defineConfig({
  resolve: {
    conditions: ['browser'],
  },
  test: {
    environment: 'jsdom',
    server: {
      deps: {
        inline: [
          '@stencil/react-output-target',
          '@lit/react',
          // plus generated wrapper package in our case
        ],
      },
    },
  },
});
```

[...]

PR #793, opened 2026-04-27
fix(react-runtime): temporary Vitest/jsdom compatibility shim for forwarded props

mhoritani:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue URL: https://github.com/stenciljs/output-targets/issues/791

In Vitest + jsdom, wrappers produced by `@stencil/react-output-target` can lose forwarded element props like `id`.

The root of it is export resolution: `@lit/react` has separate `browser` and `node` entries, and Vitest's transform context can pick the `node` entry. That `node` path is SSR-oriented and does not run the browser prop-application effect, so props classified as element props never land on the custom element in tests.

## What is the new behavior?

- Adds jsdom regression coverage for native prop forwarding (`id`) and custom element prototype property forwarding (object value path).
- Replaces runtime delegation in `packages/react/src/runtime/create-component.ts` with a temporary compatibility shim aligned with `@lit/react` browser `create-component` behavior.
- Keeps the public runtime API unchanged and leaves SSR runtime behavior untouched.

This is intentionally framed as a temporary compatibility layer, not a move away from Lit.

- [ ] Yes
- [x] No

## Other information
- I also validated with external reproductions tied to #791, including a no-workaround run for the standalone repro.
- Temporary shim exit criteria are documented in the code comments: once `@lit/react` provides a stable way to preserve browser-style prop forwarding semantics in this Vitest/jsdom scenario, the shim should be removed and direct delegation can return.
- My read is that the long-term fix likely belongs upstream in `@lit/react`. I did not want to block users on an uncertain upstream timeline, so this PR takes a short-term compatibility path.
- @johnjenkins: does this look like the right short-term direction here, or would you prefer a different approach in this repo while upstream is unresolved?

### DIFF
--- a/packages/react/src/runtime/create-component.jsdom.test.ts
+++ b/packages/react/src/runtime/create-component.jsdom.test.ts
@@ -1,0 +1,99 @@
+// @vitest-environment jsdom
+
+import React, { act } from 'react';
+import { createRoot, type Root } from 'react-dom/client';
+import { afterEach, beforeAll, describe, expect, it } from 'vitest';
+
+import { createComponent } from './create-component';
+
+const roots: Root[] = [];
+
+beforeAll(() => {
+  (globalThis as any).IS_REACT_ACT_ENVIRONMENT = true;
+});
+
+afterEach(async () => {
+  for (const root of roots.splice(0)) {
+    await act(async () => {
+      root.unmount();
+    });
+  }
+  document.body.innerHTML = '';
+});
+
+describe('createComponent (jsdom)', () => {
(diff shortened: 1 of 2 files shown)


mhoritani:
Upstream PR is now open in Lit: [https://github.com/lit/lit/pull/5325](https://github.com/lit/lit/pull/5325)

Given feedback here, I’m treating that as the primary fix path and can follow up in output-targets based on the outcome upstream.

PR #805, opened 2026-05-28
fix(react): emit components.server.ts barrel when esModules and hydrateModule are both set

Burzmalian:
## Pull request checklist

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When `reactOutputTarget` is configured with both `esModules: true` and `hydrateModule` (SSR enabled), the build:

- Writes one `${tagName}.ts` (client wrapper) per component. ✓
- Writes one `${tagName}.server.ts` (server/SSR wrapper) per component. ✓
- Writes `components.ts` as a thin re-export barrel of the per-component client files. ✓
- **Does not write `components.server.ts`.** ✗

The original (non-`esModules`) `components.server.ts` — a single file containing every server wrapper inline — is left behind from a previous build, or never created on a fresh checkout.

Consequences:

- **Silent break on fresh checkout / `git clean -fdx`** — the build succeeds with no warning, but `dist/components.server.js` is never produced. Any consumer of the SSR entry point gets `ERR_MODULE_NOT_FOUND` at runtime.
- **Tree-shaking defeated for SSR consumers** — when the original `components.server.ts` is left in place from a pre-`esModules` build, it's still the giant inline barrel, so a consumer importing one server wrapper pulls every server wrapper into the bundle, even though per-component files exist on disk.

### Root cause

In `packages/react/src/create-component-wrappers.ts`, the `esModules` branch only calls `createEsModulesComponentsFile` once, and that function is hardcoded to write `components.ts` and `./${tagName}.js` exports. There is no equivalent call for the server barrel.

Issue URL: _none filed — happy to open one if preferred._

## What is the new behavior?

- `createEsModulesComponentsFile` (`packages/react/src/create-es-modules-components-file.ts`) now accepts three new optional parameters: `filename` (defaults to `'components.ts'`), `componentSuffix` (defaults to `''`), and `serializeShadowRoot`. When `componentSuffix` is non-empty, the generated barrel additionally emits `import { type SerializeShadowRootOptions } from '@stencil/react-output-target/ssr'` and an exported `serializeShadowRoot` constant — mirroring the pattern already used in `create-stencil-react-components.ts`. Defaults preserve the previous output byte-for-byte.
- `createComponentWrappers` (`packages/react/src/create-component-wrappers.ts`) now invokes `createEsModulesComponentsFile` a second time inside the `esModules` branch, gated on `hydrateModule`, with `filename: 'components.server.ts'`, `componentSuffix: '.server'`, and the user's `serializeShadowRoot` config. The same `excludeServerSideRenderingFor` filter that already guards the per-component server files is applied to the barrel.
- Three new tests cover (1) default `serializeShadowRoot` emission, (2) JSON-stringified custom `serializeShadowRoot` emission, and (3) end-to-end behavior of `esModules: true` + `hydrateModule` + `excludeServerSideRenderingFor`. The existing `esModules: true` test (no `hydrateModule`) is unchanged, asserting backward compat.

This completes the documented `esModules` behavior — _\"better for tree-shaking\"_ — for SSR consumers.

- [ ] Yes
- [x] No

All new parameters on `createEsModulesComponentsFile` are optional with defaults that match the original behavior. The new server-barrel emission is gated on `hydrateModule`; users with `esModules: true` and no SSR are unaffected.

## Other information

Verification:

- `pnpm --filter @stencil/react-output-target test` — 44 passing (was 41).
- `pnpm --filter @stencil/react-output-target build` — clean.
- `pnpm --filter @stencil/react-output-target prettier` — all touched files unchanged.
- Verified downstream against a Next.js 15 / React 19 SSR consumer: deleting `components.server.ts` and rebuilding now regenerates it as a per-component re-export barrel with the `serializeShadowRoot` constant; the Next.js server build splits components across purpose-specific chunks instead of bundling all server wrappers into one chunk — confirming SSR tree-shaking now fires.


OS-jacobbell (approved):
Looks good!

PR #814, opened 2026-06-05
fix(angular): use per-component CustomEvent type for event outputs

### LINKED ISSUE #531: bug: event types changed from [ComponentName]CustomEvent<number> to EventName<CustomEvent<number>>
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [X] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [X] I have searched for [existing issues](https://github.com/ionic-team/stencil-ds-output-targets/issues) that already report this problem, without success.

### Stencil Version

4.22.2

### Stencil Framework Output Target

React

### Stencil Framework Output Target Version

0.7.4

### Current Behavior

Prior to upgrading to 0.7.x, in version 0.5.3, event types name were driven (I believe) from the LocalJSX (exported as JSX) of the generated components.d.ts file, which included a nomenclature of  [InterfaceName]CustomEvent.

`
declare namespace LocalJSX {
     interface MyFoo {
         "onMyEventTrigger"?: (event: MyFooCustomEvent<number>) -> void;
     }
}
`

Now with version 0.7.x, the event type name changed to `EventName<CustomEvent<number>>` within the generated component.ts.

While I understand that the event is not a complex type but I don't think it should have changed from what it was before.  This breaks any consumer of my library that I published.  What I don't understand is that the LocalJSX still has the same event type, `MyFooCustomEvent<number>` as it was in the 0.5.3 version. 

### Expected Behavior

I would have expected the event types to be same as it was before

### Steps to Reproduce

1. Created a stencil web component with an event (non complex type) using stencil version 4.20.0
2. Generate the react wrapper using react-output-target version 0.5.3
3. Create a react application that uses the web component react wrapper and handles the event using typescript
4. Upgrade the stencil library to 4.22.2 and upgrade the react-output-target to 0.7.4
5. Upgrade the react application to use the updated web component react wrapper

### Code Reproduction URL

https://github.com/foobar

### Additional Information

_No response_

ShaneK:
resolves #531 but also for Angular

## Pull request checklist

## Pull request type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Generated Angular wrappers type every event output as `EventEmitter<CustomEvent<Detail>>`. The generic `CustomEvent` types `target` as `EventTarget | null`, so when you bind a handler in a template and read `$event.target`, it doesn't narrow to the component's element under Angular strict template type checking. You have to cast it to get at element-specific members. 

Issue URL: https://github.com/stenciljs/output-targets/issues/531 is the same issue, but that one was for React. This is also happening to Angular.

## What is the new behavior?

Events are now typed with the per-component `${Component}CustomEvent<Detail>` type that Stencil generates (for example `MyComponentCustomEvent<Detail>`), whose `target` is the concrete element. Both the runtime `@Output()` emitters in `generate-angular-component.ts` and the `.d.ts` interface declarations use it, and `createComponentEventTypeImports` in `utils.ts` imports the type once per component that emits public events. This matches the React output target, which already types events this way (#716, issue #531).

- [ ] Yes
- [X] No

## Other information

This was happening as reported in #531 and fixed in https://github.com/stenciljs/output-targets/pull/716 but it was only fixed for React. This also needs to be fixed for Angular and Vue is not affected.

This may have gone unnoticed for so long because Angular before Angular 21 didn't do strict enough type checking for it to matter.


OS-jacobbell (approved):
Tested with Ionic Framework and it looks good!

PR #821, opened 2026-06-23
feat(react): reflect required props in generated React component types

Cliffback:
Props marked with `@Prop() foo!: Type` (required: true in Stencil metadata) are now emitted as a 4th generic parameter R on StencilReactComponent.

This causes TypeScript to error at compile time when a required prop is omitted in JSX, matching Angular's existing NG8008 behavior.

The R generic defaults to `never`, so existing components without required props are unaffected, no breaking change for current consumers.

Changes:
- Added R generic to StencilProps, StencilReactComponent, and both createComponent functions (client + SSR)
- Generator emits R at variable type, createComponent call site, and clientModule cast
- Type-level test suite (tsc --noEmit) validates required/optional behavior
- 4 new generator tests covering single, multiple, zero required props + SSR

Example output:
`StencilReactComponent<StnTimePickerElement, StnTimePickerEvents, Components.StnTimePicker, 'hours'>`

## Pull request checklist

## Pull request type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes

Issue URL:

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/react/src/create-stencil-react-components.ts
+++ b/packages/react/src/create-stencil-react-components.ts
@@ -177,7 +177,7 @@ import type { Components } from "${stencilPackageName}/${customElementsDir}";
 
     const componentEventNamesType = `${reactTagName}Events`;
 
-    const requiredProps = component.properties.filter((p) => p.required).map((p) => `'${p.name}'`);
+    const requiredProps = component.properties.filter((p) => p.required && !p.internal).map((p) => `'${p.name}'`);
     const requiredGeneric = requiredProps.length > 0 ? `, ${requiredProps.join(' | ')}` : '';
 
     sourceFile.addTypeAlias({
--- a/packages/react/src/create-stencil-react-components.ts
+++ b/packages/react/src/create-stencil-react-components.ts
@@ -177,6 +177,9 @@ import type { Components } from "${stencilPackageName}/${customElementsDir}";
 
     const componentEventNamesType = `${reactTagName}Events`;
 
+    const requiredProps = component.properties.filter((p) => p.required).map((p) => `'${p.name}'`);
+    const requiredGeneric = requiredProps.length > 0 ? `, ${requiredProps.join(' | ')}` : '';
+
     sourceFile.addTypeAlias({
       isExported: true,
       name: componentEventNamesType,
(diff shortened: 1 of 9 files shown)


Cliffback:
> thanks @Cliffback ! This is a non-breaking version of #788 right?
> 
> cc @ShaneK / @OS-jacobbell - check it out

Yes, that was the goal! 

We are changing our components to comply with `ban-default-true` from `@stencil/eslint-plugin`, and so need to reverse a few boolean props defaulting to `true` and realised that the `required` type checking only were propagated to the `angular-output-target` and not the `react-output-target`, so this would be very useful for us!

packages/react/src/create-stencil-react-components.ts
```diff
@@ -177,14 +177,17 @@ import type { Components } from "${stencilPackageName}/${customElementsDir}";
 
     const componentEventNamesType = `${reactTagName}Events`;
 
+    const requiredProps = component.properties.filter((p) => p.required).map((p) => `'${p.name}'`);
```
ShaneK:
```suggestion
    const requiredProps = component.properties.filter((p) => p.required && !p.internal).map((p) => `'${p.name}'`);
```
I think this needs to skip `internal` props too, the same way the Angular target already does. The Angular generator filters `!prop.internal` before it ever looks at `required` ([output-angular.ts](https://github.com/stenciljs/output-targets/blob/0432175258b5fbb85806d066a96bda4ed4fc260a/packages/angular/src/output-angular.ts#L200)), so `@internal` props never make it into the generated inputs and `@Input({ required: true })` only fires on genuinely public props. Without that filter here, a prop marked `/** @internal */` still comes through as `required: true` and is still a key of `Components.X`, so it lands in `R` and becomes required in consumer JSX.

That matters because components often have framework-assigned internal props the app author never passes. As a concrete example, Ionic core has 12 required props and 8 are `@internal` (every overlay's `overlayIndex`, plus breadcrumb's `last` and `showCollapsedIndicator`). Angular leaves those out today, so adding `&& !p.internal` here brings React in line with Angular instead of requiring props users can't even see.

ShaneK (changes requested):
Hey! I really appreciate the attempt to make this non-breaking! I did notice one issue that will be pretty impactful, but should be a simple enough fix. Other than that, this looks great to me!

packages/react/src/create-stencil-react-components.ts
ShaneK: I initially thought this was an issue with Vue's implementation too, but the way Vue's components get exported isn't the same and internal props get filtered out elsewhere, so Vue is actually safe from this at a different place.

Cliffback: Ah, I see! Thanks for the thorough explanation! We don’t use any `internal` props ourselves, so this didn’t cross my mind, and thanks!

Cliffback:
> Hey! I really appreciate the attempt to make this non-breaking! I did notice one issue that will be pretty impactful, but should be a simple enough fix. Other than that, this looks great to me!

Thank you! I applied the suggested change now

ShaneK (approved):
Awesome work, looks good to me if it's good with @johnjenkins 🚀

johnjenkins (approved):
LGTM!

Cliffback:
@johnjenkins Awesome! When can we expect the next release?

PR #826, opened 2026-06-30
refactor(react): move runtime code to build artifacts

OS-jacobbell:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [X] Refactoring (no functional changes, no api changes)
- [ ] Build related changes

The code for creating a component and the component's type are in `@stencil/react-output-target/runtime`. Component library code can be generated at compile time with one version of this package, while the user application uses a different version at runtime, leading to incompatibility that's hard to catch when changing source code for `@stencil/react-output-target`.

## What is the new behavior?

- Include type and function definitions for creating client-side Stencil React components in the generated code.
- Keep import of Lit in the npm package's runtime portion so users don't have to import it directly themselves.

- [X] Yes
- [ ] No

Since this removes code from runtime, it would make a component library generated by an older version incompatible with new runtime versions.

Both the component library and the app will need to upgrade to latest `@stencil/react-output-target`, and build artifacts will need to be regenerated.

## Other information

There's much more runtime code for SSR, so I've left that alone for now.


packages/react/src/create-stencil-react-components.ts
```diff
@@ -2,13 +2,21 @@ import type { ComponentCompilerMeta } from '@stencil/core/internal';
...
 
 interface ReactEvent {
   originalName: string;
   name: string;
   type: string;
 }
 
+function getSubstringBetween(before: string, after: string, src: string) {
+  const start = src.indexOf(before) + before.length;
+  const end = src.indexOf(after);
```
johnjenkins: nit... would never happen but still :D 

```suggestion
  const end = src.indexOf(after, start);
```

johnjenkins (approved):
I think this looks good. Just remember to include `BREAKING CHANGE: __your description__` in the extended commit message to appropriately increment the version

johnjenkins (approved):

packages/react/src/create-stencil-react-components.ts
```diff
@@ -49,17 +57,22 @@ export const createStencilReactComponents = ({
       ]
         .filter(Boolean)
         .join('\n')
-    : `import { createComponent } from '@stencil/react-output-target/runtime';`;
+    : `import { createLitComponent } from '@stencil/react-output-target/runtime';`;
+  let createComponentBody = getSubstringBetween('// @types-begin', '// @types-end', createComponentSource);
+  if (!hydrateModule) {
+    createComponentBody += getSubstringBetween('// @create-component-begin', '// @create-component-end', createComponentSource);
```
ShaneK:
```suggestion
    createComponentBody += getSubstringBetween(
      '// @create-component-begin',
      '// @create-component-end',
      createComponentSource
    );
```
This line is 128 chars, over the 120 `printWidth`, so `prettier --check` fails on it. CI didn't catch it because the `react` package has no `prettier.dry-run` script, so `lerna run prettier.dry-run` skips the whole package silently. Might be worth adding that script in a follow-up so react source actually gets format-checked.

packages/react/src/create-stencil-react-components.ts
```diff
@@ -2,13 +2,21 @@ import type { ComponentCompilerMeta } from '@stencil/core/internal';
...
+// Inlined at build time so the generated wrapper can be self-contained.
+import createComponentSource from './runtime/create-component.ts?raw';
 
 interface ReactEvent {
   originalName: string;
   name: string;
   type: string;
 }
 
+function getSubstringBetween(before: string, after: string, src: string) {
```
ShaneK, line 14: `getSubstringBetween` has no guard when a marker is missing. If `before` or `after` isn't found, `indexOf` returns `-1` and it silently returns garbage instead of erroring. I confirmed it: renaming `// @types-end` in `create-component.ts` drops the whole `StencilReactComponent` definition from the generated output with no error, it just produces a broken wrapper. I think this should throw when either marker is absent, so a bad edit fails the build loudly instead of shipping corrupt generated code.

Related: the `// @types-begin` / `// @create-component-begin` markers in `create-component.ts` are load-bearing but read like stray comments. Could you add a line there noting they're extracted at build time by this file and must not be moved, renamed, or removed? Easy for someone tidying that file to break this without realizing.

packages/react/src/runtime/index.ts
```diff
@@ -1,2 +1,2 @@
 export type { EventName, Options } from '@lit/react';
-export { createComponent, type StencilReactComponent } from './create-component.js';
+export { createComponent as createLitComponent } from '@lit/react';
```
ShaneK, line 2: This drops `createComponent` and `StencilReactComponent` from the public `@stencil/react-output-target/runtime` export. Anyone importing them directly breaks, and so does anyone who upgrades the package without regenerating their wrappers, since older generated code imports them from here. I think this needs a `packages/react/BREAKING.md` entry documenting the removed exports and the migration path (regenerate wrappers, or use `createLitComponent`). It also affects whether semantic-release cuts the major it needs, but that's the commit-message conversation, not this line.

ShaneK (changes requested):
Looking good, just noticed a few issues

packages/react/src/runtime/index.ts
ShaneK: I guess I could be convinced otherwise, but I'm worried this will break for users of reliant libraries if their version setup allows patch version updates and somebody installs and the patch version update changes but it's relying on imports that no longer exist.

I think this would even impact users of Ionic Framework (which would be something to worry about if v9 were live)

packages/react/src/create-stencil-react-components.ts
OS-jacobbell: Good catch! I went ahead and updated the prettier scripts in this PR.

packages/react/src/create-stencil-react-components.ts
OS-jacobbell: We do have test coverage for those situations, but I've added some error checking and throw-on-failure if the tags are not found. I've also added a comment explaining the begin and end markers.

packages/react/src/runtime/index.ts
OS-jacobbell: Made a major-react-2.0 branch to make it clear this will be a breaking change in a major version and updated BREAKING.md. If any other breaking changes come up, they could be included in this major-react-2.0 branch before releasing. Coinciding with Ionic Framework v9 may make the most sense.

ShaneK (approved):
Looks good to me! Awesome work! 🎉

brandyscarney (approved):

PR #838, opened 2026-08-11
chore(react): merge main into major-react-2.0

OS-jacobbell:
## Pull request checklist

## New behavior
- Use two space indentation in react build artifacts (reduce file size; make source code indentation consistent with generated code indentation).
- Updated `component-library-react` to use React 19. Since it's linked to the workspace, the `@stencil/react-output-target/runtime` does not resolve to a peer dependency, leading to mismatched React versions.
- Generalized `createLitComponent` to `createCustomElementReactComponent`.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [X] Other (please describe): big merge

- [ ] Yes
- [X] No


pnpm-lock.yaml
```diff
@@ -825,13 +840,13 @@ importers:
         version: 1.0.8(@types/react@19.2.17)
       html-react-parser:
         specifier: ^5.2.17
-        version: 5.2.17(@types/react@19.2.17)(react@19.2.7)
+        version: 5.2.17(@types/react@19.2.17)(react@18.3.1)
       react:
         specifier: ^18 || ^19
-        version: 19.2.7
+        version: 18.3.1
```
ShaneK: This looks unintentional. Both `main` and `major-react-2.0` resolve `packages/react` to `react@19.2.7`, the specifier hasn't changed, but this drops it to 18.3.1. The `@types/react` pin stays on 19.2.17, so the package type checks against 19 and runs against 18.

That puts the new `mergeClassNames` tests on React 18 here and React 19 on `main`. CI stayed green because they're pure function assertions, so the `forwardRef` path never renders.

Could you re-run `pnpm install` clean and see if this hunk goes away? I'd want it sorted before merge.

packages/react/src/runtime/create-component.ts
```diff
@@ -29,6 +30,47 @@ export type StencilReactComponent<
...
+export const mergeClassNames = (
+  currentClasses: Iterable<string>,
+  newClassName: string | undefined,
+  oldClassName: string | undefined
+): string => {
+  const incoming = new Set(splitClassName(newClassName));
+  const previous = new Set(splitClassName(oldClassName));
+  const finalClassNames: string[] = [];
+
+  for (const className of currentClasses) {
```
ShaneK, line 51: Answering the question from the description: no, not breaking for Ionic. The generated file in Ionic Framework v9 is compiled at `target: es2022` with `lib: ["dom", "es2022"]`, and I compiled this region against that config to be sure. Clean, including under `noUnusedLocals`.

For anyone on a lower target, `downlevelIteration` alone isn't enough. At `--target es5 --lib es5,dom` you get `Iterable`, `Set` and `Array.from` unresolved. Bumping `lib` to `es2015` leaves two TS2802s for the `for...of` loops, which `downlevelIteration` or `--target es2015` clears. Same reason `"target": "esnext"` had to go into the example project.

Fine to take the floor on a major, but could it get a `BREAKING.md` entry next to the `createComponent` one?

packages/react/src/runtime/create-component.ts
```diff
@@ -29,6 +30,47 @@ export type StencilReactComponent<
...
+const splitClassName = (className: string | undefined): string[] =>
+  className ? className.split(/\s+/).filter(Boolean) : [];
+
+/**
+ * Merge the app's 'className'/'class' with the classes the Stencil runtime
+ * manages on the host. '@lit/react' maps 'className' to 'class' and rewrites it
+ * wholesale on every render, which wipes runtime-added classes like 'hydrated',
+ * 'sc-*' scope classes, and state classes a design system relies on.
+ */
+export const mergeClassNames = (
```
ShaneK: This whole region is copied into the generated file, so the `export` becomes public API in every consumer's `components.ts`. The `BREAKING.md` entry covers `createComponent` and `StencilReactComponent` but not this one.

Looks like it's only exported so the unit test can reach it. Worth settling while 2.0 is unreleased, either drop the `export` and test through `createComponent`, or document it. Its neighbours `splitClassName` and `useIsomorphicLayoutEffect` are both unexported.

packages/react/src/runtime/create-component.ts
```diff
@@ -29,6 +30,47 @@ export type StencilReactComponent<
 // @types-end
 
 // @create-component-begin
+const splitClassName = (className: string | undefined): string[] =>
+  className ? className.split(/\s+/).filter(Boolean) : [];
+
+/**
+ * Merge the app's 'className'/'class' with the classes the Stencil runtime
```
ShaneK: These used backticks on `main` and became single quotes in the merge, leaving this as the only file in the package without backticks in its JSDoc. The reason is that these regions get pasted into the `dedent` template literals in `create-component-wrappers.test.ts`, where a backtick needs escaping.

Worth a line in the marker comment above, otherwise the next sync from `main` brings the backticks back and breaks the same opaque way. The codegen itself doesn't care, it interpolates a runtime string.

packages/react/src/create-component-wrappers.test.ts
```diff
@@ -52,60 +52,143 @@ type EventNames = Record<string, EventName | string>;
...
-    C = Omit<I, keyof HTMLElement>,
-    R extends keyof C = never,
+  I extends HTMLElement,
+  E extends EventNames = {},
+  C = Omit<I, keyof HTMLElement>,
+  R extends keyof C = never,
 > = React.FunctionComponent<StencilProps<I, E, C, R>>;
 
+const splitClassName = (className: string | undefined): string[] =>
+  className ? className.split(/\\s+/).filter(Boolean) : [];
```
ShaneK: All three `toEqual(dedent...)` expectations now carry a full copy of the runtime, hand escaped, so the source's `/\s+/` shows up here as `/\\s+/`. That's most of the added lines.

Since every future expectation will carry the runtime too, could these move to `toMatchSnapshot()`? There's precedent in `packages/ssr/test/__snapshots__/`, and it turns a runtime edit into a `-u` instead of three re-escapes. Up to you, the current form does fail loudly.

.github/workflows/ci.yml
```diff
@@ -15,13 +15,13 @@ jobs:
     timeout-minutes: 15
     steps:
       - name: 📥 Checkout Code
-        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - name: 📦 Setup pnpm
         uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6.0.9
```
ShaneK: This is from a snapshot of `main` rather than its tip, so two dependabot bumps got missed. The `pnpm/action-setup` pin is still v6.0.9 against v6.0.10 on `main`, and `actions/stale` is still v10 against v11. Both were on `main` before this commit.

Next sync picks them up, so no big deal, but a rebase would sort it now. The checkout and setup-node bumps that did come across are consistent.

ShaneK (changes requested):
I noticed a few issues with this merge, most probably just bad merge artifacts

.github/workflows/ci.yml
OS-jacobbell: I merged in latest main

packages/react/src/runtime/create-component.ts
OS-jacobbell: Added backticks back in. The snapshot tests automatically escape them, so no longer any need to replace with single quotes.

packages/react/src/runtime/create-component.ts
OS-jacobbell: I moved all the export statements to outside the copy markers.

packages/react/src/runtime/create-component.ts
OS-jacobbell: Added to `BREAKING.md`.

packages/react/src/create-component-wrappers.test.ts
OS-jacobbell: Switched these to snapshot tests.

pnpm-lock.yaml
OS-jacobbell: I have no idea what I'm doing with pnpm lock files, so I reverted it to `main`'s.

packages/react/src/runtime/create-component.ts
```diff
@@ -51,11 +93,56 @@ export const createComponent = <
...
+  });
+
+  WrappedComponent.displayName = options.displayName ?? finalTagName;
+
+  return WrappedComponent as unknown as StencilReactComponent<I, E, C, R>;
 };
 // @create-component-end
+
+export type { StencilReactComponent };
+export { mergeClassNames, createComponent };
```
ShaneK: I think this went further than I meant. I only had `mergeClassNames` in mind, but `createComponent` and `StencilReactComponent` moved out too, and those two need to stay exported since the generated file is where consumers get them now.

Importing either from the generated package fails with "has no exported member", where the same import compiles fine against the generated file from before this branch. That also makes the `BREAKING.md` line above wrong, since it tells apps they can import these from their components package.

CI won't catch it, since nothing in the repo imports either symbol and the barrel is just `export * from './components.js'`.

Could you put `export` back on the two declarations inside the markers and leave this line as `export { mergeClassNames };`? The helpers can stay unexported inside the markers since the copied `createComponent` calls them, and `mergeClassNames` only needs the export down here for the unit test.

packages/react/BREAKING.md
```diff
@@ -3,6 +3,7 @@
 ### Breaking Changes
 
 - The `createComponent` funtion and `StencilReactComponent` type have been moved from `@stencil/runtime` to generated build artifacts. Apps that imported these directly can now import them from their Stencil React components package. Stencil React components generated with `@stencil/react-output-target` v1 will not be compatible with `@stencil/react-output-target` v2 runtime.
+- Building the generated React component library requires tergeting es2015 at minimum. Projects can set `"target": "es2015"` or higher in their `tsconfig.json` files.
```
ShaneK:
```suggestion
- Building the generated React component library requires targeting es2015 at minimum. Projects can set `"target": "es2015"` or higher in their `tsconfig.json` files.
```

Just the typo. The es2015 floor itself is right, that target compiles the generated file clean.

Nit on the line above while you're here: "funtion" should be "function", and `@stencil/runtime` should be `@stencil/react-output-target/runtime`.

ShaneK (changes requested):
Awesome work! Two minor corrections, though

PR #827, opened 2026-07-01
fix(react): merge className with runtime-managed host classes

ShaneK:
Resolves #485

## Pull request checklist

Please check if your PR fulfills the following requirements:

- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [X] Build (`npm run build`) was run locally for affected output targets
- [X] Tests (`npm test`) were run locally and passed
- [X] Prettier (`npm run prettier`) was run locally and passed

## Pull request type

Please check the type of change your PR introduces:

- [X] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

`@lit/react` maps React's `className` prop to the host's `class` attribute and rewrites it wholesale on every render. That wipes the classes the Stencil runtime manages on the host: `hydrated`, `sc-*` scope classes, mode classes, and state classes like `interactive`. So when an app drives `className` (for example a form library setting Ionic's `ion-invalid` / `ion-touched` validation classes), those app classes replace the runtime's classes instead of coexisting with them, and the component loses styling and hydration state.

Issue URL:
- https://github.com/stenciljs/output-targets/issues/485
- https://github.com/ionic-team/ionic-framework/issues/31237

## What is the new behavior?

`createComponent` now wraps the `@lit/react` component and withholds `className`/`class` so React never writes the `class` attribute directly. It reconciles the app's class value against the live host in a layout effect using a new `mergeClassNames` helper. The merge keeps runtime-managed classes the app never set, appends new app classes, and removes only the app classes that were dropped since the previous render (tracked with a `previousClassName` ref). It uses an isomorphic layout effect so it falls back to `useEffect` during server rendering and doesn't warn.

## Does this introduce a breaking change?

- [ ] Yes
- [X] No

PR #840, opened 2026-08-21
fix(vue): remove conditional classes that are no longer bound

### LINKED ISSUE #835: bug: [vue] reactive class bindings are never removed - wrapper accumulates classes and re-applies stale ones
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/ionic-team/stencil/blob/main/.github/CONTRIBUTING.md).
- [x] I agree to follow the [Code of Conduct](https://github.com/ionic-team/stencil/blob/main/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/stenciljs/output-targets/issues) that already report this problem, without success.

### Framework Output Target

Vue (@stencil/vue-output-target)

### Output Target Version

0.13.2

### Stencil Version

4.43.5

### Reproduction URL

https://stackblitz.com/edit/wjcrly83?file=src%2Finput.vue

### Describe the Bug

Sandbox is not opening. So I have provided the stackblitz link.
https://stackblitz.com/edit/wjcrly83?file=src%2Finput.vue

When a reactive `:class` binding on a wrapped component becomes inactive, the class is not
removed from the element. The wrapper re-applies it imperatively, so the component stays stuck
in a stale visual state.

Root cause is in `packages/vue/src/runtime.ts`.

**1. The class set is add-only.** `classes` is created once in `setup()` and every render only
ever adds to it - it is never reconciled against the current `attrs.class`:

```ts
const classes = new Set(getComponentClasses(attrs.class));

return () => {
  getComponentClasses(attrs.class).forEach((value) => {
    classes.add(value); // add-only, nothing is ever removed
  });
```

**2. Stale classes are written back to the DOM imperatively**, bypassing Vue's class diffing,
in `syncElementClasses()`:

```ts
componentClasses.forEach((c) => {
  if (!!c && !element.classList.contains(c)) {
    element.classList.add(c); // re-adds classes the binding no longer wants
  }
});
```

**3. Vue never owns those classes.** `syncElementClasses()` excludes `componentClasses` from the
`class` value it returns for the vnode, so Vue has no mechanism to remove them.

There is a secondary effect that makes this confusing to debug. The returned vnode class is
`element.classList` minus the set. On the render where the binding flips off, that returned
string *changes*, so Vue runs `patchClass`, overwrites `className`, and the stale class
disappears — it looks correct. On any later render where the returned string is unchanged, Vue
skips `patchClass` entirely (it only patches when `prev !== next`), so the imperative
`classList.add()` from step 2 sticks and the class visibly comes back.

That is why the class appears to toggle off correctly once, then reappears on a later,
unrelated update.

Minor, related: `getComponentClasses('')` returns `['']`, so an empty class binding inserts an
empty string into the set.

I intend to submit a PR for this — proposed fix is in Additional Context.

### Steps to Reproduce

https://stackblitz.com/edit/wjcrly83?file=src%2Finput.vue

1. Open the stackblitz link and focus the input.
2. Type 6 characters (e.g. `output`), then click outside to blur.
   → `isInvalid` becomes `true`, `is-invalid` is added, the input is outlined red. Correct.
3. Press Backspace once (5 characters left).
   → `isInvalid` becomes `false`, the red outline disappears. Still correct.
4. Press Backspace once more (4 characters left).
   → `isInvalid` is still `false` and the Vue-rendered "Too long" text stays hidden, but the
     red outline **returns** - `is-invalid` is back on the element.

The `class attribute mutations` log on the page shows the stale class being re-added:

```
class -> "is-invalid"   <- step 2, correct
class -> ""             <- step 3, correct
class -> "is-invalid"   <- step 4, BUG: re-added while isInvalid === false
```

The contrast in step 4 is the key signal: the `v-if` text is rendered by Vue and behaves
correctly, while the class is managed imperatively by the wrapper and is stale.

### Expected Behavior

The wrapper should reconcile against the current value of `attrs.class` and remove classes that
are no longer present, so reactive class bindings behave exactly as they do on a native element.

For `<ExampleInput :class="{ 'is-invalid': isInvalid }" />`, once `isInvalid` is `false` the
`is-invalid` class should be removed from the DOM and stay removed across subsequent re-renders.

### Additional Context

We currently fixed this as a `pnpm patch`. It makes the set reconcile instead of
accumulate:

```js
const nextClasses = new Set(getComponentClasses(attrs.class));
classes.forEach((value) => {
  if (!nextClasses.has(value)) {
    containerRef.value?.classList.remove(value);
  }
});
classes.clear();
nextClasses.forEach((value) => classes.add(value));
```

and filters empty entries so `classList.remove('')` cannot throw:

```js
const getComponentClasses = (classes) => (classes?.split(' ') || []).filter(Boolean);
```

Happy to raise a PR with this approach if it looks reasonable.

### Logs

```shell-session

```

### LINKED ISSUE #31392: bug: a removed conditional `:class` reappears on Ionic components in Vue
## Ionic Framework Version

- [x] v9.x
- [x] Nightly

(Not present in v8.8.19.)

## Current Behavior

In `@ionic/vue` 9, a class supplied by a conditional Vue binding can reappear on an Ionic component after it has correctly been removed.

`syncElementClasses` (new in v9) imperatively re-adds every class the component has *ever* received via `attrs.class` on every render. Vue's own `patchClass` normally overwrites the element's class afterwards — but Vue skips that patch when the newly computed class string is identical to the previous render's. On any such render the imperative re-add survives, and the stale class is back.

That makes it look intermittent: the class is dropped correctly on the render where the class string changes, then returns on the next render where it does not.

```vue
<ion-item :class="{ 'is-active': selected === id }">…</ion-item>
```

Select a → b and `a` correctly loses `is-active`. Select b → c and `a` gets `is-active` back, so two items are active at once.

This is a behaviour change from v8, where the same template worked correctly.

## Expected Behavior

A class supplied by a conditional binding should stay removed once the binding stops producing it, as in v8 and as for non-Ionic elements.

## Steps to Reproduce

1. `ionic start stickyClass blank --type=vue`
2. Install `@ionic/vue@9.0.0` / `@ionic/vue-router@9.0.0` / `@ionic/core@9.0.0`.
3. Render three `ion-item`s driven by one selection ref — three are needed, because the bug only shows on the *second* deselection render:

```js
import { IonicVue, IonItem, IonList } from '@ionic/vue'
import { createApp, h, ref } from 'vue'
import '@ionic/vue/css/core.css'

const IDS = ['a', 'b', 'c']
const selected = ref('a')

const App = {
  setup () {
    return () => h(IonList, null, {
      default: () => IDS.map((id) => h(
        IonItem,
        { id: `item-${id}`, class: { 'is-active': selected.value === id } },
        { default: () => `item ${id}` },
      )),
    })
  },
}

createApp(App).use(IonicVue).mount('#app')

const settle = () => new Promise((r) => setTimeout(r, 250))
const snapshot = () => IDS
  .map((id) => `${id}:${document.getElementById(`item-${id}`).classList.contains('is-active') ? 'ACTIVE' : '-'}`)
  .join('  ')

await settle()
console.log('mounted (a):     ', snapshot())
selected.value = 'b'; await settle()
console.log('after selecting b:', snapshot())
selected.value = 'c'; await settle()
console.log('after selecting c:', snapshot())
```

4. Read the console.

**Actual on 9.0.0:**

```
mounted (a):       a:ACTIVE  b:-  c:-
after selecting b: a:-  b:ACTIVE  c:-        <- correctly removed
after selecting c: a:ACTIVE  b:-  c:ACTIVE   <- 'a' is active again
```

**Expected (and what 8.8.19 produces):** exactly one active item at every step.

Only the `@ionic/vue` version changes between the two runs; everything else is identical.

## Code Reproduction URL

No Stackblitz link, sorry — but the reproduction is fully inlined in **Steps to Reproduce** above and needs nothing beyond a stock `ionic start … --type=vue` app with that snippet pasted into a single page. The behaviour also flips purely on the `@ionic/vue` version, with no other change. Happy to put it on Stackblitz if that would help triage.

## Ionic Info

```
Ionic:

   Ionic CLI       : 7.2.1
   Ionic Framework : @ionic/vue 9.0.0

Capacitor:

   Capacitor CLI      : 8.5.0
   @capacitor/android : 8.5.0
   @capacitor/core    : 8.5.0
   @capacitor/ios     : 8.5.0

System:

   NodeJS : v26.7.0
   npm    : 11.19.0
   OS     : macOS
```

### Cause

`defineContainer` keeps a `Set` of the classes Vue has supplied, and that Set is only ever added to (`@ionic/vue/dist/index.js`):

```js
const classes = new Set(getComponentClasses(attrs.class));   // setup
…
return () => {
    getComponentClasses(attrs.class).forEach((value) => {
        classes.add(value);                                   // render — never removes
    });
    …
    const propsToAdd = {
        …
        class: syncElementClasses(containerRef, classes),
    };
};
```

`syncElementClasses` is new in v9 and force-adds every member of that Set to the host element on each render:

```js
const syncElementClasses = (ref, componentClasses, defaultClasses = []) => {
    if (ref?.value) {
        const element = ref.value;
        // makes sure vue classes are on the actual element
        componentClasses.forEach((c) => {
            if (!!c && !element.classList.contains(c)) {
                element.classList.add(c);
            }
        });
    }
    return [...Array.from(ref.value?.classList || []), ...defaultClasses].filter((c, i, self) => {
        return !componentClasses.has(c) && self.indexOf(c) === i;
    });
};
```

In v8 the equivalent helper (`getElementClasses`) never touched the DOM — it only computed a list that was handed back to Vue, so Vue's own class patching handled both add and remove. In v9 the imperative add has no counterpart. It is normally masked because `patchClass` rewrites `el.className` wholesale straight afterwards, but Vue skips that when the computed class string is unchanged from the previous render — and then the re-added stale class survives.

### Suggested fix

Track the classes from the current render rather than accumulating, and remove the ones that are no longer bound:

```js
const nextClasses = new Set(getComponentClasses(attrs.class));
if (containerRef.value) {
    classes.forEach((value) => {
        if (!!value && !nextClasses.has(value)) {
            containerRef.value.classList.remove(value);
        }
    });
}
classes.clear();
nextClasses.forEach((value) => classes.add(value));
```

### Impact

In our app this affects 68 conditional `:class` bindings on Ionic components across 51 files, so it cannot be worked around in templates. It failed 7 of 40 E2E shards, typically as duplicated "active"/"selected" state — e.g. two side-menu `ion-item`s both carrying `is_active` after navigating through three tabs, and an `ion-item` that still had `is_selected`. Because it depends on two consecutive renders producing the same class string, it reads as flakiness rather than a deterministic bug, which made it expensive to track down.

We are currently shipping the patch above as a local `pnpm patch`.

Found while upgrading one app from 8.8.19 to 9.0.0; two more regressions from the same release: [#31393](https://github.com/ionic-team/ionic-framework/issues/31393) (`getComponentClasses` whitespace split, same helper) and [#31394](https://github.com/ionic-team/ionic-framework/issues/31394) (`ion-radio` value reflect).

ShaneK:
## Pull request checklist

## Pull request type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently, `defineContainer` accumulates every class Vue supplies into a `Set` and never removes any, so `syncElementClasses` keeps re-adding classes the binding has dropped. Usually Vue's `patchClass` overwrites `className` wholesale and cleans up incidentally, but on a render where that string is unchanged Vue skips the patch and the re-added class sticks. A `v-for` of slot-bearing components hits it easily, since unstable slots re-render every item on each selection, including ones whose class binding didn't change.

Issue URL: resolves #835

## What is the new behavior?

The whole class cycle now lives in `syncElementClasses`: remove what Vue supplied last render and no longer supplies, add the current classes, record them for next time. Removal has to precede the return value, which is the element's classes minus the ones Vue supplied - a stale class left on would be reported back as runtime-managed and rewritten on every later render. This fixes ionic-team/ionic-framework#31392 in Ionic Framework. 

The tokenizer changed too, and the two aren't separable. Splitting on a single space leaves whitespace inside the tokens of a multi-line class attribute, which `classList` rejects (ionic-team/ionic-framework#31393). It now splits on runs of ASCII whitespace and drops empty tokens, which the removal loop needs since `classList.remove('')` throws.

The regex is `[ \t\n\f\r]+` rather than `\s+` because `\s` matches ten characters `DOMTokenList` doesn't split on. A value containing U+00A0 would record `foo` and `bar` when the element's only real token is `foo\u00a0bar`, so nothing could ever remove it. React's `splitClassName` can be looser because it rewrites the whole attribute; this runtime mutates individual tokens, so it has to match the DOM.

Seven unit tests cover both issues, plus an e2e test in `vue-app` with a `ReactiveClassTest.vue` fixture modelled on `HydratedClassTest.vue`.

- [ ] Yes
- [X] No

## Other information

The React target had the same bug, fixed in #827. The tokenizer now matches React's `splitClassName` and the reconcile computes the same result as `mergeClassNames`

### DIFF
--- a/packages/vue/src/runtime.ts
+++ b/packages/vue/src/runtime.ts
@@ -37,27 +37,61 @@ interface NavManager<T = any> {
   navigate: (options: T) => void;
 }
 
+/**
+ * Split a class attribute value into tokens the way `DOMTokenList` does, on runs of ASCII
+ * whitespace. A single-space split leaves whitespace inside the tokens of a multi-line class
+ * attribute, and `classList` rejects those. The token set has to match the DOM's exactly: one
+ * we record that the DOM never produced can never be removed again.
+ *
+ * React's `splitClassName` gets away with a looser `\s+` because it rewrites the whole `class`
+ * attribute, so its own tokens become the truth. We mutate individual tokens on an element the
+ * DOM already tokenized, so the two regexes are deliberately different.
+ */
 const getComponentClasses = (classes: unknown) => {
-  return (classes as string)?.split(' ') || [];
+  return ((classes as string)?.split(/[ \t\n\f\r]+/) || []).filter((className) => className.length > 0);
 };
 
+/**
+ * Sync the classes Vue supplied onto the host element, and return the ones it didn't supply
+ * (Stencil's `hydrated` and `sc-*` scope classes, plus anything added imperatively) so Vue's
+ * class patch doesn't wipe them. Vue merges that return value back with `attrs.class`.
+ *
+ * We read `previousVueClasses` for last render's Vue classes, then rewrite it with this
+ * render's.
+ *
+ * Classes Vue has stopped supplying have to come off the element before that return value is
+ * computed, or a stale one gets reported back as a class Vue never supplied and rewritten on
+ * every later render. Deleting the removal loop reintroduces issue #835.
+ */
 const syncElementClasses = (
   ref: Ref<HTMLElement | undefined>,
-  componentClasses: Set<string>,
+  previousVueClasses: Set<string>,
+  vueClasses: unknown,
   defaultClasses: string[] = []
 ) => {
+  const nextClasses = new Set(getComponentClasses(vueClasses));
+
   if (ref?.value) {
     const element = ref.value;
+    // drops the vue classes that are no longer bound
+    previousVueClasses.forEach((c) => {
+      if (!nextClasses.has(c)) {
+        element.classList.remove(c);
+      }
+    });
     // makes sure vue classes are on the actual element
-    componentClasses.forEach((c) => {
-      if (!!c && !element.classList.contains(c)) {
+    nextClasses.forEach((c) => {
+      if (!element.classList.contains(c)) {
         element.classList.add(c);
       }
     });
   }
 
+  previousVueClasses.clear();
+  nextClasses.forEach((c) => previousVueClasses.add(c));
+
   return [...Array.from(ref.value?.classList || []), ...defaultClasses].filter((c: string, i, self) => {
-    return !componentClasses.has(c) && self.indexOf(c) === i;
+    return !nextClasses.has(c) && self.indexOf(c) === i;
   });
 };
 
@@ -114,7 +148,7 @@ export const defineContainer = <Props, VModelType = string | number | boolean>(
     (props, { attrs, slots, emit }) => {
       let modelPropValue = modelProp ? props[modelProp as keyof InputProps<VModelType>] : undefined;
       const containerRef = ref<HTMLElement>();
-      const classes = new Set(getComponentClasses(attrs.class));
+      const previousVueClasses = new Set<string>();
 
       onMounted(() => {
         /**
@@ -210,10 +244,6 @@ export const defineContainer = <Props, VModelType = string | number | boolean>(
       return () => {
         modelPropValue = props[modelProp as keyof InputProps<VModelType>];
 
-        getComponentClasses(attrs.class).forEach((value) => {
-          classes.add(value);
-        });
-
         // @ts-expect-error
         const oldClick = props.onClick;
         const handleClick = (ev: Event) => {
@@ -227,7 +257,7 @@ export const defineContainer = <Props, VModelType = string | number | boolean>(
 
         const propsToAdd: Record<string, unknown> = {
           ref: containerRef,
-          class: syncElementClasses(containerRef, classes),
+          class: syncElementClasses(containerRef, previousVueClasses, attrs.class),
           onClick: handleClick,
         };
 
--- a/example-project/vue-app/src/components/ReactiveClassTest.vue
+++ b/example-project/vue-app/src/components/ReactiveClassTest.vue
@@ -1,0 +1,62 @@
+<script setup lang="ts">
+import { MyList, MyListItem } from 'component-library-vue'
+import { ref } from 'vue'
+
+const ids = ['a', 'b', 'c']
+const selected = ref('a')
+</script>
+
+<template>
+  <div class="reactive-class-test">
+    <h3>Reactive Class Removal Test (Issue #835)</h3>
+    <p>Selecting an item should leave exactly one item with the "is-active" class.</p>
+
+    <MyList>
+      <MyListItem
+        v-for="id in ids"
+        :key="id"
+        :data-testid="`reactive-item-${id}`"
+        :class="{ 'is-active': selected === id }"
+      >
+        item {{ id }}{{ selected === id ? ' (selected)' : '' }}
+      </MyListItem>
+    </MyList>
+
+    <div class="controls">
+      <button
+        v-for="id in ids"
+        :key="`select-${id}`"
+        :data-testid="`select-item-${id}-btn`"
+        @click="selected = id"
+      >
+        Select {{ id }}
+      </button>
+      <p data-testid="selected-item">
+        selected: <strong>{{ selected }}</strong>
+      </p>
+    </div>
+  </div>
+</template>
+
+<style scoped>
+.reactive-class-test {
+  margin: 20px 0;
+  padding: 20px;
+  border: 2px solid #ccc;
+  border-radius: 8px;
+}
+
+.controls {
+  margin-top: 15px;
+}
+
+button {
+  padding: 10px 20px;
+  font-size: 16px;
+  cursor: pointer;
+}
+
+:deep(.is-active) {
+  outline: 3px solid orange;
+}
+</style>
--- a/example-project/vue-app/src/views/BasicComposition.vue
+++ b/example-project/vue-app/src/views/BasicComposition.vue
@@ -37,6 +37,7 @@
     Check the DOM: the tag should be &lt;v1-my-transform-test&gt; instead of &lt;my-transform-test&gt;
   </p>
   <HydratedClassTest />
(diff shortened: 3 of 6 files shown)


packages/vue/src/runtime.ts
```diff
@@ -37,27 +37,61 @@ interface NavManager<T = any> {
   navigate: (options: T) => void;
 }
 
+/**
+ * Split a class attribute value into tokens the way `DOMTokenList` does, on runs of ASCII
+ * whitespace. A single-space split leaves whitespace inside the tokens of a multi-line class
+ * attribute, and `classList` rejects those. The token set has to match the DOM's exactly: one
+ * we record that the DOM never produced can never be removed again.
+ *
+ * React's `splitClassName` gets away with a looser `\s+` because it rewrites the whole `class`
```
thetaPC, line 46: Should this be Vue?

```suggestion
 * Vue's `splitClassName` gets away with a looser `\s+` because it rewrites the whole `class`
```

thetaPC (approved):
LGTM

packages/vue/src/runtime.ts
ShaneK: No, this comment is highlighting why we're doing it differently here. This is shorthand comparing React's version of this same code to Vue's by saying "React does this a different way because [...]" so if somebody comes along and says "Why doesn't this work the same way?" this comment lets them know

OS-jacobbell (approved):
Looks good!

PR #842, opened 2026-08-27
feat(angular): support setting boolean props by attribute presence

thetaPC:
## Pull request checklist

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Boolean props have to be bound explicitly. A bare attribute resolves to the empty string, so `strictTemplates` reports `Type 'string' is not assignable to type 'boolean'`.

Issue URL: https://github.com/ionic-team/ionic-framework/issues/30822

## What is the new behavior?

A new opt-in `booleanAttributes` option declares boolean props with an Angular input transform, so they can be set by attribute presence:

```html
<my-component disabled></my-component>

<!-- rather than -->
<my-component [disabled]="true"></my-component>
```

- Boolean props are emitted using Angular's `inputs` metadata object form, e.g. `{ name: 'disabled', transform: nullableBooleanAttribute }`. `ProxyCmp` still proxies the property, so runtime plumbing is untouched.
- Applies to both single file and `esModules` generation.
- Non-boolean props, virtual props, and the `@angular/core` imports are unchanged.
- Requires Angular **16.1** or later, which is when input transforms were added.
- Enabled in the example project, with an e2e test covering attribute presence, `disabled="false"`, and property bindings.

Off by default. With the option unset, generated output is identical to today.

- [ ] Yes
- [x] No

## Other information

The transform is a small `nullableBooleanAttribute` helper rather than Angular's own `booleanAttribute`, and that is the one decision worth reviewing closely.

Angular's `booleanAttribute` coerces both `null` and `undefined` to `false`. Components commonly compare boolean props by strict identity, so that coercion would change the behavior of bindings that work today. In Ionic this includes `focusTrap === false` on `ion-modal` and `ion-popover`, and `swipeGesture === false` on `ion-segment-view`. Both props default to `true`, so `<ion-modal [focusTrap]="obs$ | async">` would go from keeping the focus trap active to silently disabling it, because the `async` pipe emits `null` before its first value. Form control values do the same. The helper only coerces strings and passes `null` and `undefined` through, so the sole behavior change is the intended one.

Two notes for reviewers:

- The helper is declared as a `function` rather than an arrow constant. Angular resolves input transforms statically in partial compilation and rejects the arrow form with `NG1010: Input transform must be a function`.
- It lives in `angular-component-lib/utils.ts` and deliberately does not import from `@angular/core`, so consumers on Angular versions without `booleanAttribute` are unaffected by the file being generated.

I also confirmed the transform survives the library boundary rather than only working on in-app source: the built `.d.ts` carries an `ngAcceptInputType_*` entry per transformed prop, which is what widens the accepted type for apps consuming the published package.


OS-jacobbell (approved):

packages/angular/angular-component-lib/utils.ts
```diff
@@ -44,6 +44,41 @@ export const defineCustomElement = (tagName: string, customElement: any) => {
...
+ * first emission and from form control values, so coercing them would change the behavior of
+ * bindings that work today.
+ *
+ * This is implemented here rather than imported from `@angular/core` so that consumers on
+ * Angular versions without `booleanAttribute` are unaffected by this file being generated.
+ *
+ * Declared as a function rather than an arrow constant because Angular has to resolve input
+ * transforms statically when compiling a library in partial compilation mode.
+ */
+export function nullableBooleanAttribute(value: unknown): boolean | null | undefined {
```
ShaneK:
```suggestion
export function nullableBooleanAttribute(
  value: boolean | string | null | undefined
): boolean | null | undefined {
```

Using `unknown` here makes Angular emit `ngAcceptInputType_disabled: unknown`, so `[disabled]="42"` compiles under `strictTemplates`. Could we narrow it to `boolean | string | null | undefined`?

Stencil already handles the documented values, and with this signature the bare attribute, `disabled="false"`, and boolean bindings still compile while invalid bindings error again. All 72 Angular unit tests pass too.

example-project/ng-app/tests/test.e2e.ts
```diff
@@ -46,6 +46,30 @@ describe('Stencil Angular Integration', () => {
     await expect(counter.$('span')).toHaveText('2278');
   });
 
+  it('should set boolean props from attribute presence', async () => {
```
ShaneK: I think this passes with `booleanAttributes` turned off too. With no transforms generated, the proxy and Stencil element still produce all four values asserted here because Stencil already coerces boolean properties.

Could this be a build-time template test with `inlineProperties` enabled instead? That would cover the `strictTemplates` behavior the option changes.

packages/angular/README.md
```diff
@@ -57,3 +57,43 @@ export const config: Config = {
...
+| `booleanAttributes`    | When `true`, boolean properties are declared with an Angular input transform so they can be set by attribute presence, e.g. `<my-component disabled>` instead of `<my-component [disabled]="true">`. Requires Angular 16.1 or later. Defaults to `false`. Refer to [Boolean attributes](#boolean-attributes).                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      |
```
ShaneK: Could this mention that the `strictTemplates` benefit also requires `inlineProperties`? Without it Angular doesn't type-check these input properties, and the example project currently enables only `booleanAttributes`.

The Angular 16.1 requirement is also below the package's supported v19+ floor. One other limitation I found: Stencil resolves something like `boolean | 'auto'` to compiler type `any`, so it gets no transform and a bare attribute still produces `TS2322`. Might be worth noting that next to the virtual-property limitation.

packages/angular/src/output-angular.ts
```diff
@@ -154,6 +188,12 @@ export function generateProxies(
    */
   const componentLibImports = ['ProxyCmp'];
 
+  const booleanAttributes = outputTarget.booleanAttributes === true;
+
+  if (usesInputTransform(components, booleanAttributes)) {
+    componentLibImports.push('nullableBooleanAttribute');
```
ShaneK: Could this use `INPUT_TRANSFORM_FUNCTION` instead? Renaming that constant currently changes the emitted transform reference but leaves this import as `nullableBooleanAttribute`, so the Angular build fails with `TS2552` and `NG1010`. This file already imports from `generate-angular-component`, so using the constant here and in the `esModules` path should keep both sides together.

ShaneK (approved):
Solid addition overall. I left a few comments around preserving template type checking, tightening the docs and tests, and keeping the generated import in sync.

packages/angular/angular-component-lib/utils.ts
thetaPC: https://github.com/stenciljs/output-targets/pull/842/changes/0ba7974161237f743c2906320debdd22fa3d563a

example-project/ng-app/tests/test.e2e.ts
thetaPC: https://github.com/stenciljs/output-targets/pull/842/changes/db00b3b995849499eaabce62797f0139e726f8f2

packages/angular/README.md
thetaPC: https://github.com/stenciljs/output-targets/pull/842/changes/6ad77130e870120af9d453a1a559a8685a41a26c

packages/angular/src/output-angular.ts
thetaPC: https://github.com/stenciljs/output-targets/pull/842/changes/a3a4fb25358dd45a184c5589c8414eacf49f2f03

packages/angular/README.md
```diff
@@ -57,3 +57,69 @@ export const config: Config = {
...
+```
+
+Without `inlineProperties` the generated wrappers declare no typed members for their inputs, so
+Angular does not check these bindings at all and there is no `TS2322` for the transform to
+resolve. The runtime is unaffected either way, because Stencil already coerces boolean attributes
+on the element itself.
+
+### Limitations
+
+Two kinds of property are never transformed, and a bare attribute on them still fails to compile:
```
ShaneK: Nit: I don't think this consequence holds for both bullets, since the two behave differently.

Virtual properties never make it into `inlineComponentProps`, so they get no class member. On `my-popover` the `mode` virtualProp sits in the `inputs` array with no setter, while every real prop there has one. Angular skips bindings to a member-less input entirely, so a bare attribute compiles and works, since Stencil coerces it anyway.

An `any`-collapsed type does get a setter, so `''` is assignable and it compiles, and then the component receives `''` instead of `true`. There's a live one already, since `my-popover` declares `event` as `any`.

So `boolean | 'auto'` is the only case that errors. The other two are silent, which needs different advice than binding explicitly. Worth splitting the consequence per bullet.

One other thing while you're in here, the `booleanAttributes` JSDoc only documents the union limitation and drops the virtual-property carve-out this section has.

packages/angular/src/output-angular.ts
```diff
@@ -15,12 +15,50 @@ import {
...
+ * whether the transform needs to be imported.
+ *
+ * @param components The components in the generated file.
+ * @param booleanAttributes Whether boolean properties should declare an input transform.
+ * @returns `true` when at least one property is transformed.
+ */
+const usesInputTransform = (components: readonly ComponentCompilerMeta[], booleanAttributes: boolean) =>
+  booleanAttributes &&
+  components.some((cmpMeta) =>
+    (cmpMeta.properties ?? []).filter(filterInternalProps).some((p) => p.type === 'boolean')
```
ShaneK: Nit: this re-derives the eligibility rule that `mapInputProp` already applies, down to its own `filterInternalProps` pass. They agree today and I couldn't find a config where they diverge, since both only read `booleanAttributes` and the `boolean` type check over the same filtered set.

If that rule ever widens though, changing one and not the other emits a proxies file referencing `nullableBooleanAttribute` with no import, which dies in the consumer's build rather than here. None of the new tests would catch it either, because they all drive both sides from the same `type: 'boolean'` fixture. Could you pull the predicate into one function they both call?

The hoist also dropped the "Ensure that virtual properties has required as false" comment, so that carve-out now only exists as the missing second argument at the two `virtualProperties` call sites. Reads like someone forgot it rather than meant it. A JSDoc line would keep the intent where the function is. No worries if you'd rather leave it.

packages/angular/angular-component-lib/utils.ts
```diff
@@ -44,6 +44,41 @@ export const defineCustomElement = (tagName: string, customElement: any) => {
...
+ * first emission and from form control values, so coercing them would change the behavior of
+ * bindings that work today.
+ *
+ * This is implemented here rather than imported from `@angular/core` so that consumers on
+ * Angular versions without `booleanAttribute` are unaffected by this file being generated.
+ *
+ * Declared as a function rather than an arrow constant because Angular has to resolve input
+ * transforms statically when compiling a library in partial compilation mode.
+ */
+export function nullableBooleanAttribute(value: boolean | string | null | undefined): boolean | null | undefined {
```
ShaneK: Nit: nothing really pins down what the narrowing bought us. Angular builds `ngAcceptInputType_*` from this parameter type, so that narrowing is what makes `[disabled]="42"` error again, but `ng build` only ever proves that valid bindings compile. If this widened back to `unknown` the whole matrix would stay green.

A type-level test would pin it cheaply and needs no build artifact, something like `expectTypeOf` on the parameter plus a `@ts-expect-error` on a number call, since both halves fail if it widens. Would need `--typecheck` on the package's vitest run. Up to you!

example-project/component-library/stencil.config.ts
```diff
@@ -66,6 +66,8 @@ export const config: Config = {
       directivesProxyFile: '../component-library-angular/projects/library/src/directives/proxies.ts',
       valueAccessorConfigs: angularValueAccessorBindings,
       transformTag: true,
+      booleanAttributes: true,
+      inlineProperties: true,
```
ShaneK, line 77: Nit, and not asking for a change here. Flipping this means the library build, the ng-app build and the wdio suite all now run against `inlineProperties`, so the default shape that most consumers actually ship is only covered by unit tests from here. I checked and the flag really is load-bearing for the build-time test, because without it there's no `TS2322` for the transform to resolve, so I think it's the right trade.

Might be worth a line in the PR description saying the example is deliberately non-default, just so nobody takes it as the recommended config. The option is still marked Experimental and there's no note anywhere about what would promote it.

ShaneK (approved):
Looks good to me, great work! All four of my earlier comments are handled. Left a few nits, but none of them are blocking 🚀

packages/angular/README.md
thetaPC: https://github.com/stenciljs/output-targets/pull/842/changes/c0a1cbc5e1c3da72296cfc50889dbffe9bb91ff4

packages/angular/src/output-angular.ts
thetaPC: https://github.com/stenciljs/output-targets/pull/842/changes/084d1e89796c3c055d3bdee573f3cccbe46e554f

packages/angular/angular-component-lib/utils.ts
thetaPC: https://github.com/stenciljs/output-targets/pull/842/changes/6313dd3b8ee3443193c1a525759464e14947fd5b and https://github.com/stenciljs/output-targets/pull/842/changes/a646ab5ee584bf630411e6f8a7b3d2fe2cc465b9

example-project/component-library/stencil.config.ts
thetaPC: https://github.com/stenciljs/output-targets/pull/842/changes/3e57c7372d8b96a904198f5b8560b300fe28df1b

packages/angular/tsconfig.spec.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "noEmit": true,
+    "types": ["node"]
+  },
+  "include": ["src", "tests", "angular-component-lib"]
```
ShaneK:
```suggestion
  "include": ["src", "tests", "angular-component-lib/boolean-attribute.ts"]
```

My fault for suggesting the typecheck without thinking about the include. Pulling in the whole directory picks up `utils.ts`, whose `rxjs` import doesn't resolve since `rxjs` isn't a dep here, so vitest exits 1. All 74 tests pass and there are no type errors, which makes it a confusing one to read. That's what the angular job is failing on for both macOS and Windows.

Narrowing it to the one file exits 0 with everything still green, and it keeps `boolean-attribute.ts` clear of `rxjs` the way you intended. I checked it doesn't weaken the type test either, widening the parameter back to `unknown` still fails both halves.

packages/angular/src/types.ts
```diff
@@ -95,6 +95,39 @@ export interface OutputTargetAngular {
...
+   * <!-- otherwise the value has to be bound explicitly -->
+   * <my-component [disabled]="true"></my-component>
+   * ```
+   *
+   * Without a transform Angular resolves the bare attribute to the empty string and reports
+   * `Type 'string' is not assignable to type 'boolean'` under `strictTemplates`.
+   *
+   * This only has a visible effect where Angular type-checks the wrapper's inputs, which also
+   * requires `inlineProperties`. Without it the generated wrappers declare no typed members for
+   * their inputs, so these bindings are never checked. The runtime is unaffected either way,
```
ShaneK: I don't think the runtime is unaffected here. Angular applies the transform from the component def's input map, and that doesn't care whether there's a declared class member, so it runs with `inlineProperties` off too. It replaces Stencil's coercion rather than deferring to it.

Stencil and the transform agree on strings, booleans, `null` and `undefined`, so most of the surface really is identical. Falsy non-strings are where they split, so something like `[disabled]="items.length"` disables on a non-empty list today and would always be true now.

With `inlineProperties` on, `ngAcceptInputType_*` catches that at compile time. On its own it doesn't, and this doc and the README section both say `booleanAttributes` alone is fine, so the flip is silent. Could we soften the sentence in both places? I don't think the coercion needs to change.

packages/angular/src/generate-angular-component.ts
```diff
@@ -34,20 +34,38 @@ function createPropertyDeclaration(
 }
 
 /**
- * Creates a formatted inputs text with required declaration.
+ * The transform function to reference in generated code, exported from
+ * `angular-component-lib/utils`.
```
ShaneK:
```suggestion
 * The transform function to reference in generated code, exported from
 * `angular-component-lib/boolean-attribute`.
```

Looks like this survived the commit that moved the transform to its own module, since both emit sites in `output-angular.ts` point at `boolean-attribute`.

Same commit left the test filenames behind. Calling them `angular-component-lib.test.ts` and `.test-d.ts` was right when the helper was in `utils.ts`, but every other test file here is named for the module it covers, and that name is the obvious one for a future `angular-component-lib/utils.ts` test. Renaming to `boolean-attribute.*` is safe, vitest still picks both up including the typecheck one. Up to you though.

ShaneK (approved):
All the nits from last round are handled, great work! I'm approving instead of requesting changes since the red angular job stops this merging anyway, but that one does need the typecheck include narrowed before you can merge. The rest is nits

packages/angular/tsconfig.spec.json
thetaPC: https://github.com/stenciljs/output-targets/pull/842/changes/ffeced8d785e710e590b354fd9156158c0d273c7

packages/angular/src/types.ts
thetaPC: https://github.com/stenciljs/output-targets/pull/842/changes/bd87af86fb0c6d69f7539c3e3d29d02345bb4f50

packages/angular/src/generate-angular-component.ts
thetaPC: https://github.com/stenciljs/output-targets/commit/ffeced8d785e710e590b354fd9156158c0d273c7

ShaneK (approved):

